PR #50, opened 2022-03-31
deploy user impression flow

lfishhead:
this PR Intends to deploy the user impression data flow that Ian started.  The initial version of the flow relied on data being regularly dumped from dbt to parquet files on S3.  The flow then loaded the data to a feature group `user-impressions`.  

since the initial work, we have new models of the per=user impression age of recommended items in `first_impressed_age` on snowflake.  This allows us to now deploy this flow to update the feature group using the modeled data, and in turn build out the impression-capping ranker that will use this data to filter items from slates that users have seen previously.

## choices

1. the impression age and impression count are currently used as filters to create the list of items that shouldn't be shown to each user.  the flow script has parameters that specify default values for the impression age (6 days) and impression count (12)
2. `first_impressed_age` includes a `time_added` field indicating when the list for the corresponding `user_id` was updated.  I am only updating the feature group if `time_added` is within the last two weeks.  This assumes everything will be new for users who haven't impressed items for two weeks, and should overwrite the row for that `user_id` in any event.
3. we are still using `resolved_id` here although we can convert this feature group to `content_id` or `url` at any time.  This is just in keeping with the existing schema and also the fact that the recommendation-api broadly speaking knows nothing of `content_id`

## TIL

that [recit](https://github.com/Pocket/recit-ralph/blob/623ab3a7fefdfb9544139f0db790fa52e41e8da9/application.py#L145) is actually using this data currently for generating personalized recommendations in the web home experiment:


src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,85 @@
...
+SELECT
+    u.USER_ID,
+    p.HAPPENED_AT AS UPDATED_AT,
+    LISTAGG(DISTINCT p.RESOLVED_ID, ',') AS RESOLVED_IDS
+FROM prep as p
+JOIN ANALYTICS.DBT.USERS as u
+  ON p.HASHED_USER_ID = u.HASHED_USER_ID
+WHERE (p.max_impression_age > %(MAX_AGE)s) OR (p.total_imprs > %(MAX_IMPRS)s)
+GROUP BY 1, 2
+"""
```
mmiermans, line 41: This query took a minute to execute. Might be fine in practice? If not, would be good to get G's eyes on it, given that he's our SQL expert.

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,83 @@
...
     boto_session = boto3.Session()
     feature_store_session = Session(boto_session=boto_session,
                                     sagemaker_client=boto_session.client(service_name='sagemaker'),
                                     sagemaker_featurestore_runtime_client=boto_session.client(service_name='sagemaker-featurestore-runtime'))
     feature_group = FeatureGroup(name=feature_group_name, sagemaker_session=feature_store_session)
     feature_group.ingest(df, max_workers=4, max_processes=4, wait=True)
 
+# Schedule to run every hour
+if config.ENVIRONMENT == config.ENV_PROD:
+    schedule = IntervalSchedule(interval=datetime.timedelta(hours=1))
```
mmiermans, line 65:
```suggestion
    schedule = IntervalSchedule(interval=datetime.timedelta(days=1))
```

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,83 @@
...
-    load_featue_group.map(df=xdfs, feature_group_name=unmapped(feature_group))
+    max_impr_age = Parameter("max impression age", default=6)
+    max_impr_count = Parameter("max impression count", default=12)
+
+    snowflake_result = PocketSnowflakeQuery()(
+        query=BASE_QUERY,
+        data={"MAX_AGE": max_impr_age, "MAX_IMPRS": max_impr_count}
+    )
+
+    xdf = transform_user_impressions_df(snowflake_result)
```
mmiermans, line 79: Rename to a longer name, to be more explicit.

src/flows/recommendation_api/user_impressions_flow.py
lfishhead: @gkatre can you have a look at the SQL in this flow?  I would like to run the flow daily and only update the feature store for `user_id`s that have new impression data

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+  
+SELECT
+    u.USER_ID,
+    p.HAPPENED_AT AS UPDATED_AT,
+    LISTAGG(DISTINCT p.RESOLVED_ID, ',') AS RESOLVED_IDS
+FROM prep as p
+JOIN ANALYTICS.DBT.USERS as u
+  ON p.HASHED_USER_ID = u.HASHED_USER_ID
+WHERE ((p.max_impression_age > %(MAX_AGE)s) OR (p.total_imprs > %(MAX_IMPRS)s))
+  AND CURRENT_DATE = p.HAPPEEND_AT
```
mmiermans, line 40:
```suggestion
  AND CURRENT_DATE = p.HAPPENED_AT
```
^ Want to run the flow one more time to ensure the query is working?

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+  JOIN ANALYTICS.DBT.CONTENT AS c
+    ON c.CONTENT_ID = a.CONTENT_ID
+  WHERE CURRENT_DATE - DATE(a.TIME_ADDED) <= 1  
+  GROUP BY 1, 2, 3
+  )
+  
+SELECT
+    u.USER_ID,
+    p.HAPPENED_AT AS UPDATED_AT,
+    LISTAGG(DISTINCT p.RESOLVED_ID, ',') AS RESOLVED_IDS
```
mmiermans, line 35:
```suggestion
    ARRAY_AGG(DISTINCT p.RESOLVED_ID) AS RESOLVED_IDS
```
[array_agg](https://docs.snowflake.com/en/sql-reference/functions/array_agg.html) surrounds the values by brackets.

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+    max_impr_age = Parameter("max impression age", default=6)
+    max_impr_count = Parameter("max impression count", default=12)
+
+    snowflake_result = PocketSnowflakeQuery()(
+        query=BASE_QUERY,
+        data={"MAX_AGE": max_impr_age, "MAX_IMPRS": max_impr_count}
+    )
+
+    xformed_df = transform_user_impressions_df(snowflake_result)
+    load_feature_group(df=xformed_df, feature_group_name=feature_group)
```
mmiermans, line 81:
```suggestion
    transformed_df = transform_user_impressions_df(snowflake_result)
    load_feature_group(df=transformed_df, feature_group_name=feature_group)
```
I got what this meant, but I usually like to write names out, unless the shortened name is used frequently, to lower the barrier for other devs.

mmiermans (commented):
Found a typo, and I have a couple minor suggestions. Overall looks good.

src/flows/recommendation_api/user_impressions_flow.py
lfishhead: nice

lfishhead: this inserts newline characters to make an array it seems: `[id1\n, id2\n, ..., id33\n]` .  i'm going back to listagg

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+prep as (
+  SELECT
+    a.HASHED_USER_ID,
+    a.CONTENT_ID,
+    c.RESOLVED_ID,
+    MAX(a.IMPRESSION_AGE) as max_impression_age,
+    SUM(a.IMPRESSION_COUNT) as total_imprs,
+    MAX(a.HAPPENED_AT_DAY) as HAPPENED_AT
+  FROM ANALYTICS.DBT.FIRST_IMPRESSED_AGE as a
+  JOIN ANALYTICS.DBT.CONTENT AS c
```
gkatre, line 26: Is it possible to use only `content_id` instead of `resolved_id`? This join seems unnecessary.
If we are going to need `resolved_id`s then we should consider putting those in the models (example: `FIRST_IMPRESSED_AGE`) we use and avoid the joins

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+    ON c.CONTENT_ID = a.CONTENT_ID
+  WHERE CURRENT_DATE - DATE(a.TIME_ADDED) <= 1
```
gkatre, line 28:
```suggestion
  WHERE a.TIME_ADDED >= CURRENT_DATE -  1 
```
^^ This may be a little better on query performance

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -1,49 +1,84 @@
...
+  WHERE CURRENT_DATE - DATE(a.TIME_ADDED) <= 1  
+  GROUP BY 1, 2, 3
+  )
+  
+SELECT
+    u.USER_ID,
+    p.HAPPENED_AT AS UPDATED_AT,
+    LISTAGG(DISTINCT p.RESOLVED_ID, ',') AS RESOLVED_IDS
+FROM prep as p
+JOIN ANALYTICS.DBT.USERS as u
```
gkatre, line 37: Can we use the `HASHED_USER_ID` instead of the `USER_ID` and avoid the join?

gkatre (commented):
@lfishhead I have left a few comments on the SQL

src/flows/recommendation_api/user_impressions_flow.py
lfishhead: this choice is related to the recommendation-api which doesn't currently know about content_id since it is older than that

src/flows/recommendation_api/user_impressions_flow.py
lfishhead: this is similar.  we are using the user_id in the personalized rankers because that's what we get with the request to recommendation-api

src/flows/recommendation_api/user_impressions_flow.py
jeshuaborges: @mmiermans has been working on this. We may be closer than you think.

gkatre (approved):
LGTM!

src/flows/recommendation_api/user_impressions_flow.py
gkatre: If the joins are causing a performance issue, we could include the `user_id` in the `FIRST_IMPRESSED_AGE` model

PR #69, opened 2022-05-06
Flow readme

mmiermans:
# Goal
Allow us to add documentation to flows that's accessible in the Prefect Cloud UI.

All you need to do is create a Markdown file with the same filename and path as the flow. This documentation is then automatically deployed in CodePipeline along with the flow itself.

## Implementation decisions
I added a warning against editing the readme in the Prefect UI, to prevent people from loosing their changes, as continuous deployment will overwrite them on each deployment.

Screenshot from 2022-05-09 12-16-04

### DIFF
--- a/deploy/register_flows.py
+++ b/deploy/register_flows.py
@@ -1,14 +1,19 @@
 #!/usr/bin/env python
+import json
 import os
 from os import environ
 from typing import Callable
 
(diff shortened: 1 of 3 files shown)


jeshuaborges (approved):
This is great.

PR #72, opened 2022-05-10
Schedule prospecting jobs with Prefect

jeshuaborges:
We are running orchestration challenges between Airflow, DBT and Metaflow while generating prospects for curation tools. It looks like we can fairly easily fix the DBT and Metaflow challenges. This is a first stab at an approach. This is not tested and is presented for feedback.

Done:
* Moved Braze jobs to its own flow and schedule
* Made DBT modeling tasks run nightly and trigger AWS step functions

Remaining:
* [x] [Setup Snowflake Stages](https://github.com/Pocket/data-warehouse/pull/64)
  * Create database
  * Create tables
* [ ] Create Snowflake IAM role `SnowflakeS3ReaderRole`
* [ ] Create S3 bucket ~~pocket-primary-db~~ `pocket-snowflake-raw`
* [ ] Create RDS IAM role
* [ ] [Create RDS user grant](https://pocket.slack.com/archives/C01CLPFMJJ1/p1652396879936789?thread_ts=1652389861.623299&cid=C01CLPFMJJ1) 
* [ ] Add all prospecting step functions
* [ ] Ensure all necessary DBT models were in the Prefect scheduled DBT Cloud job
* [ ] DBT Cloud
  * Move the braze event modeling task to its own DBT Cloud job
  * Add models to `dbt_orchestration_flow`: `stg_curated_feed_items`, `stg_curated_feed_prospects`, `CONTENT_ENGAGEMENT_BY_DAY_USER_API_ID`, `stg_tile_source`
* [ ] Metaflow - Remove prospecting tasks schedules
* [ ] DBT Cloud


src/flows/prospecting/orchestration_flow.py
```diff
@@ -0,0 +1,96 @@
...
+    return StepActivate(state_machine_arn=arn, execution_name=execution_name)
+
+
+@task()
+def prospect():
+    """
+    Run all prospecting step functions at the same time.
+    """
+    timestamp = datetime.now().strftime("%Y%m%d %H%M%S")
+    return execute_step.map(PROSPECTING_STEP_FUNCTIONS, unmapped(timestamp))
```
mmiermans, line 84: I'm curious whether this would cause the StepActivate task to run, because you're not calling `.run()` on the tasks. Easy enough to confirm by running the flow locally.

src/flows/braze/update_flow.py
```diff
@@ -359,7 +369,9 @@ def mask_email_domain_outside_production(rows: List[Dict], email_column='EMAIL')
     return rows
 
 
-with Flow(FLOW_NAME, executor=LocalDaskExecutor(), result=get_s3_result()) as flow:
+with Flow(FLOW_NAME, schedule=schedule, executor=LocalDaskExecutor(), result=get_s3_result()) as flow:
+    dbt_run_result = dbt.DbtCloudRunJob()(cause=FLOW_NAME, job_id=DBT_CLOUD_BRAZE_JOB_ID, wait_for_job_run_completion=True)
```
mmiermans, line 373: On line 391, on the `PocketSnowflakeQuery`, you'll want to add:
```python
.set_upstream(
        dbt_run_result,  # Run the Snowflake query after the Dbt job finishes
    )
```
(See `create_email_aliases` below for an example.)

mmiermans (commented):
Looks awesome! Huge improvement! :rocket:

.aws/src/config/index.ts
```diff
@@ -43,6 +43,9 @@ const prefect = {
       'GCE_KEY',
       'BRAZE_API_KEY',
       'BRAZE_REST_ENDPOINT',
+      'AURORA_DB_USER',
+      'AURORA_DB_PASSWORD',
+      'AURORA_DB_HOST',
```
mmiermans, line 48: Could you add these to the ReadMe under 'Initial Deployment'? There's a [running list parameters](https://github.com/Pocket/data-flows#2-aws-ssm-parameter-store) we need to take if we'd ever need to deploy this from scratch.

jeshuaborges: Yeah if we have secrets to production resources in Dev we should delete them and replace them with dummy values. The dev AWS account is not a safe place for this information.

mmiermans: Thanks for the call-out. I've removed Dbt and Snowflake credentials from Pocket-Dev.

src/flows/prospecting/orchestration_flow.py
jeshuaborges: Yeah I am not sure but looking at the docs this is the way they said to do it. so... we will see. :)

src/flows/braze/update_flow.py
jeshuaborges: Done

gkatre, line 373: Perhaps we could add the number of retries with retry delay.
It could be passed into the constructor of the task. https://docs.prefect.io/api/latest/core/task.html#task-2
as:
```
dbt.DbtCloudRunJob(max_retries=3, retry_delay=datetime.timedelta(seconds=2))(cause=FLOW_NAME, job_id=DBT_CLOUD_BRAZE_JOB_ID, wait_for_job_run_completion=True)
```
^^ untested code

jeshuaborges: Yeah, I added that to the prospecting flow but didn't add it here because this job was running every 10 minutes anyway. I could go either way for this flow.

src/flows/prospecting/orchestration_flow.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+with Flow(FLOW_NAME, schedule=schedule, executor=LocalDaskExecutor()) as flow:
+    extract_result = extract()
+    transform_result = transform(extract_result)
+    load_result = load(transform_result)
```
gkatre, line 90: We may not need the `load` task because the `extract` appears to be doing the extract+load into Snowflake?

src/flows/prospecting/orchestration_flow.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@task()
+def extract():
+    return execute_flow.map([f.FLOW_NAME for f in IMPORT_MYSQL_FLOWS])
+
+
+@task()
+def load():
+    return execute_flow.map([f.FLOW_NAME for f in IMPORT_MYSQL_FLOWS])
```
gkatre, line 68:
```suggestion
    return execute_flow.map([f.FLOW_NAME for f in DATA_FLOWS])
```

src/utils/config.py
```diff
@@ -26,3 +28,11 @@
 
 BRAZE_API_KEY=os.getenv('BRAZE_API_KEY')
 BRAZE_REST_ENDPOINT=os.getenv('BRAZE_REST_ENDPOINT')
+
+AURORA_DB_NAME='readitla_ril-tmp'
+AURORA_DB_HOST = os.environ.get('AURORA_DB_HOST')
+AURORA_DB_USER = os.environ.get('AURORA_DB_USER')
+AURORA_DB_PASSWORD = os.environ.get('AURORA_DB_PASSWORD')
```
gkatre, line 35: We usually add new configs in the `.env.example` file to describe the variables

src/flows/braze/update_flow.py
mmiermans: I don't feel strongly either. Fine to leave as-is. Good that G noticed this though.

PR #74, opened 2022-05-25
Setting flow schedules for Prod only deployments

gkatre:
## Goal

To implement Prefect Flow schedules to be set for the Production environment only for the FF New Tab Flow. Previously, the schedules were being deployed to both the deployment environments (Production and Development)

## Implementation Decisions

While the required change to the FF New Tab Flow was being done, an improvement to create a common function `utils.flow.get_interval_schedule` has been created that applies the schedule for the PROD environment only and is now being used in each Flow where schedule is required.

### DIFF
--- a/src/flows/dbt_orchestration_flow.py
+++ b/src/flows/dbt_orchestration_flow.py
@@ -1,15 +1,12 @@
-import datetime
-
 from prefect import Flow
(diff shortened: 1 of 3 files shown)


mmiermans (approved):
We :pear:'ed on this.

PR #75, opened 2022-05-26
Delete snowplow events for users requesting account deletion

jeshuaborges:
## Goal

Delete raw snowplow events for any user who has requested deleting their account in the last 90 days.

## Implementation Decisions
 
* Assume that we only need to delete events that were delivered from a logged-in user.
* Only worrying about users deleted within the last 90 days. This gives us at least 2 tries at a monthly cadence to delete the event records.  This ensures that if `STG_ACCOUNT_DELETIONS` grow substantially we are keeping the number of records we are processing to a minimum.

## Deployment steps
- [ ] Test Snowflake delete command.
- [ ] Ensure the Prefect Snowflake user account has delete permissions on the snowplow events schema.

## References

Documentation:
* [Project doc](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2628780051/Data+Retention+and+Deletion)


src/flows/data_retention/user_snowplow_events_flow.py
```diff
@@ -0,0 +1,28 @@
...
+from prefect.schedules import Schedule
+from prefect.schedules.clocks import CronClock
+
+from api_clients.pocket_snowflake_query import PocketSnowflakeQuery
+from utils.flow import get_flow_name
+
+FLOW_NAME = get_flow_name(__file__)
+
+# Run first day of month at midnight UTC.
+SCHEDULE = Schedule(clocks=[CronClock("0 0 1 * *")])
```
kirill-demtchouk, line 16: This timing should be OK, but worth calling out that the Drop Deleted User Account Data dbt job [runs](https://cloud.getdbt.com/#/accounts/4171/projects/7569/jobs/17774/settings/) at 12pm UTC on the 1st of the month and the last step is to truncate the `STG_ACCOUNT_DELETIONS` table.

What this means is that between when that job ends, and the next time we run our nightly workload, the `STG_ACCOUNT_DELETIONS` table is empty and we want to avoid running this flow during that time window.

kirill-demtchouk: Sure, we could definitely go that route!

src/flows/data_retention/user_snowplow_events_flow.py
```diff
@@ -0,0 +1,31 @@
...
+
+DBT_MODEL_CLEANUP_JOB_ID = 17774
+
+DELETE_SQL = """
+DELETE FROM "SNOWPLOW"."ATOMIC"."EVENTS" as e
+USING "ANALYTICS"."DBT_STAGING"."STG_ACCOUNT_DELETIONS" as d 
+WHERE e.CONTEXTS_COM_POCKET_USER_1[0]:hashed_user_id = d.hashed_user_id
+"""
```
kennylong, line 25: I always get nervous about `DELETE` statements on tables like this, and check with a `SELECT` before hand, but alas, the `USING` part doesn't work with a `SELECT`:

```
-- DELETE 
SELECT *
FROM "SNOWPLOW"."ATOMIC"."EVENTS" as e
USING "ANALYTICS"."DBT_STAGING"."STG_ACCOUNT_DELETIONS" as d 
WHERE e.CONTEXTS_COM_POCKET_USER_1[0]:hashed_user_id = d.hashed_user_id
AND d.happened_at > DATEADD("day", -90, NOW())
```
This is simple and looks good to me, but would love a second 👀 + ✅ from @kirill-demtchouk , so we don't have to test out the limits of Snowflake's [Time Travel](https://docs.snowflake.com/en/user-guide/data-availability.html) feature for the first time.

kennylong:
@jeshuaborges Will there be an associated task to run this once without the "last 90 days" limit?

src/flows/data_retention/user_snowplow_events_flow.py
```diff
@@ -0,0 +1,31 @@
...
+
+with Flow(FLOW_NAME, schedule=SCHEDULE) as flow:
+    snowflake = PocketSnowflakeQuery()
```
kennylong, line 28: @jeshuaborges Would it be possible to create a special Snowflake user/connection for this, so that all our Prefect development doesn't happen with a connection that can delete from one of our most important tables? Or some other mitigation.

src/flows/data_retention/user_snowplow_events_flow.py
kirill-demtchouk: If we want to be extra-sure we're deleting all user data, I think we may want to scan a few more entities/fields here besides `hashed_user_id`. These are all of the entities including hashed/de-coded user ID.

Some of these may be redundant, but I figure we want to be thorough. I also added a type cast to `hashed_user_id`:
```
WHERE (e.CONTEXTS_COM_POCKET_USER_1[0]:hashed_user_id::string = d.hashed_user_id
  or e.CONTEXTS_COM_POCKET_USER_1[0]:user_id::integer = d.user_id
  or e.CONTEXTS_COM_POCKET_ACCOUNT_1[0]:hashed_user_id::string = d.hashed_user_id
  or e.CONTEXTS_COM_POCKET_ACCOUNT_1[0]:user_id::integer = d.user_id
  or e.CONTEXTS_COM_POCKET_PAYMENT_SUBSCRIPTION_1[0]:user_id::integer = d.user_id
)

Comment (kennylong) [reply src/flows/data_retention/user_snowplow_events_flow.py:25]: I'm not against this, and appreciate being thorough. One thing that @jeshuaborges and I discussed yesterday was that if `hashed_user_id` is not available in event but it is related to a user account, that is either an incomplete/incorrect implementation of the event, or an implementation bug. And that we should fix the event, not account for the incorrect implementation in this process. That makes sense to me from a cleanliness/responsibility standpoint, but in order to go that route I think we would want a way to either enforce that with code or some sort of alerting/monitoring thing. I'm personally not sure what to do in this case, for our H1 milestone.

Comment (kirill-demtchouk) [reply src/flows/data_retention/user_snowplow_events_flow.py:25]: In a casual scan of our dbt models, I think the vast majority have `hashed_user_id` filled-in via the `user` entity. I was being conservative with that above suggestion haha. I agree that we could probably get away with filtering on just `hashed_user_id` in the `user` entity.

Comment (jeshuaborges) [reply src/flows/data_retention/user_snowplow_events_flow.py:25]: Lots in this thread and a lot of good comments.

> Re the query. 

100% we should be very confident that our query is doing exactly what we want before we deploy this. I would expect that once we give the thumbs up for this approach we do the select then delete approach as outlined above, in a console. If all looks good then we should move forward with taking this to prod.

> Re the possible missed context and larger where statement:

+1 to what @kennylong said. I really like the idea of having very specific requirements for our clients. If the clients are not meeting the requirement let's sort it out with the client. What I don't want to do is to keep trying to find the place that user ids may be in every event or entity remember to update our deletion processes when we make changes. If it turns out that it is unrealistic for clients to fulfill this expectation then we can follow up with another approach for standardizing the deletes.

File: src/flows/data_retention/user_snowplow_events_flow.py (hunk 3 above)
Comment (jeshuaborges) [reply src/flows/data_retention/user_snowplow_events_flow.py:28]: @kennylong great suggestion. I was also thinking that this would be a good idea. We can absolutely do this.

File: src/flows/data_retention/user_snowplow_events_flow.py [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+from prefect.tasks.dbt import DbtCloudRunJob
+from prefect.tasks.snowflake import SnowflakeQuery
+
+from utils.config import SNOWFLAKE_DATA_RETENTION_CONNECTION_DICT
+from utils.flow import get_flow_name
+
+FLOW_NAME = get_flow_name(__file__)
+
+# Run first day of month at midnight UTC.
+SCHEDULE = Schedule(clocks=[CronClock("0 0 1 * *")])
```
gkatre, line 18:
```suggestion
if ENVIRONMENT == ENV_PROD:
    SCHEDULE = Schedule(clocks=[CronClock("0 0 1 * *")])
else:
    SCHEDULE = None
```

jeshuaborges:
We need to put account deletions into a table/schema/db that is protected and will never be cleaned up. This is because we need to know which records we always need to delete.

Tables:
* deleted_users (user_id, hashed_user_id) - appended from `STG_ACCOUNT_DELETIONS`.
* deleted_emails(email) - __confirmed__ email addresses for users in deleted_users.

PR #77, opened 2022-05-26
[HS-34] Create a simple corpus candidates set

mmiermans:
## Goal
Store a set of Corpus Items in the Feature Store that we can use to show people during onboarding. We include topics to allow RecommendationAPI to filter these candidates based on the topics that users will select.

## Corpus Candidate Sets feature group
This feature group stores a set of corpus items based on a unique identifier, similar to the `RECAPI-Dev-CandidateSets` DynamoDB table, but with CorpusItems instead of legacy Pocket items.

| Feature Name | Feature Type                                              |
|--------------|-----------------------------------------------------------|
| id           | String                                                    |
| unloaded_at  | String                                                    |
| corpus_items | String (JSON list of Corpus Items, with 'id' and 'topic') |

## Implementation Decisions
- Include any approved items, not just scheduled ones, to get more stories for each topic.
- For now, sort simply on recency.

## Deployment
- [x] Create feature group `development-corpus-candidate-sets-v1`
- [x] Create feature group `production-corpus-candidate-sets-v1`

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-34

### DIFF
--- a/src/flows/recommendation_api/curated_corpus_candidates_flow.py
+++ b/src/flows/recommendation_api/curated_corpus_candidates_flow.py
@@ -1,0 +1,74 @@
+from typing import Dict, Sequence
+import json
+import datetime
+
+from prefect import task, Flow, Parameter, unmapped
+from prefect.tasks.aws.s3 import S3List
+import pandas as pd
+import boto3
+from sagemaker.feature_store.feature_group import FeatureGroup, FeatureValue
+from sagemaker.session import Session
+
+from api_clients.pocket_snowflake_query import PocketSnowflakeQuery, OutputType
+from utils import config
+from utils.flow import get_flow_name
+
+FLOW_NAME = get_flow_name(__file__)
+
+CURATED_CORPUS_CANDIDATE_SET_ID = 'deea0f06-9dc9-44a5-b864-fea4a4d0beb7'
+
+# Export scheduled corpus items
+EXPORT_SQL = """
+SELECT APPROVED_CORPUS_ITEM_EXTERNAL_ID
+FROM "SCHEDULED_CORPUS_ITEMS"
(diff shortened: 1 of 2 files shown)


src/flows/recommendation_api/curated_corpus_candidates_flow.py
```diff
@@ -0,0 +1,76 @@
...
+
+FLOW_NAME = get_flow_name(__file__)
+
+NEW_TAB_EN_US_CORPUS_CANDIDATE_SET_ID = 'deea0f06-9dc9-44a5-b864-fea4a4d0beb7'
+
+# Export approved corpus items by language and recency
+EXPORT_CORPUS_ITEMS_SQL = """
+SELECT
+    APPROVED_CORPUS_ITEM_EXTERNAL_ID as ID,
+    TOPIC
```
lfishhead, line 24: `CORPUS_REVIEW_STATUS` can be `corpus` or `recommended` in this model, this might be useful information to preserve

src/flows/recommendation_api/curated_corpus_candidates_flow.py
```diff
@@ -0,0 +1,76 @@
...
+FROM "APPROVED_CORPUS_ITEMS"
+WHERE LANGUAGE = %(language)s
+AND REVIEWED_CORPUS_ITEM_CREATED_AT BETWEEN DATEADD(day, %(scheduled_at_start_day)s, CURRENT_TIMESTAMP) AND CURRENT_TIMESTAMP
+ORDER BY REVIEWED_CORPUS_ITEM_CREATED_AT DESC
```
lfishhead, line 28: may need to filter on `SCHEDULED_SURFACE_ID`

lfishhead (approved):

PR #83, opened 2022-06-15
Flow to capture the deleted user accounts in a separate protected table

gkatre:
## Goal

Flow to capture the account deletions into separate protected tables that will never be cleaned up. 
To perform account deletions (user_ids as well as emails) across all data sets, we need to maintain the list of accounts that are to be deleted.

## Implementation Decisions

The user_ids and emails are captured in seperate tables so that we are not able to establish the relation between the two sets, while still being able to keep an independent list for the purposes of executing the deletions


mmiermans (approved):
:+1: Looks good.

PR #84, opened 2022-06-16
Remove Snowplow Events for Deleted accounts

gkatre:
## Goal

Remove Snowplow raw events for deleted accounts

## Implementation Decisions

- Changing the flow structure to pickup the Snowflake statements from a `sql` file instead of a string. This will:
   - improve maintainability with separation of flow code and sqls
   - make it easier to add multiple sql statements in a single file to run as a single task within the flow
- Files added:
  - `deleted_account_users.sql` : For maintaining a  permanent record of deleted account user_ids/hadhed_user_ids
  - `deleted_account_emails.sql`: For maintaining a  permanent record of emails for the deleted accounts
  - `delete_snowplow_events.sql`: SQL for removing the Snowplow Raw events for the deleted accounts
  -  `delete_raw_user_rows.sql`: SQL for removing the Raw events from Firehose events for the deleted accounts
  - `create_test_data.sql`: Generate test data for testing the script and permissions

## TODOs:
- [x] Create a protected DB and Schema for storing the list of account deleted tables
- [x] Change the `delete_snowplow_events.sql` script for the account deleted tables to point to this new protected DB/Schema
- [x] Add Parameter Store values for Snowplow DB/Schema, Raw data DB/Schema
- [x] Test permissions for account data deletions setup in Snowflake: https://github.com/Pocket/data-warehouse/pull/68
- [x] Dry run: Successful
- [x] Test run on a small test data (15 day data) 

## Notes:
- The time taken to delete the raw events on the 15 day test data appear to be substantial, especially for the Snowplow Raw Events
- For the monthly run, we may need to consider a separate Snowflake WH to run this job


gkatre:
@kennylong / @kirill-demtchouk I am continuing from where Jesh left off and I have this PR for deleting Snowflake raw events. (I am still working on the other raw event deletion part...will have to create test data to verify the delete scripts etc)
Would appreciate if you could review this PR mainly for the SQL I have used. Thanks!

src/flows/data_retention/delete_snowplow_events.sql
```diff
@@ -0,0 +1,5 @@
+-- Removes Snowplow Events (SNOWPLOW.ATOMIC.EVENTS) for deleted accounts
+delete from EVENTS as e
+using development.gaurang_data_retention_deletion.DELETED_USERS as d
```
kirill-demtchouk, line 3: Looks reasonable to me! Is this table meant to eventually serve as the permanent record of user IDs that have been deleted?

kirill-demtchouk (approved):

src/flows/data_retention/delete_snowplow_events.sql
gkatre: Yes. The `DELETED_USERS` and the `DELETED_EMAILS` table will hold the permanent record of user IDs and email IDs. They will exist in a separate protected DB/Schema (yet to be created and protected from deletions etc)

I will update the confluence doc with the information

src/utils/config.py
```diff
@@ -29,12 +30,16 @@
...
+    "role": os.getenv("SNOWFLAKE_DATA_RETENTION_ROLE"),
+    "database": os.getenv('SNOWFLAKE_DATA_RETENTION_DB'),
+    "schema": os.getenv('SNOWFLAKE_DATA_RETENTION_SCHEMA')
 }
 
-SNOWFLAKE_DATA_RETENTION_DB = os.getenv('SNOWFLAKE_DATA_RETENTION_DB')
-SNOWFLAKE_DATA_RETENTION_SCHEMA = os.getenv('SNOWFLAKE_DATA_RETENTION_SCHEMA')
+SNOWFLAKE_SNOWPLOW_DB = os.getenv('SNOWFLAKE_SNOWPLOW_DB')
+SNOWFLAKE_SNOWPLOW_SCHEMA = os.getenv('SNOWFLAKE_SNOWPLOW_SCHEMA')
+SNOWFLAKE_RAWDATA_DB = os.getenv('SNOWFLAKE_RAWDATA_DB')
```
jeshuaborges, line 44: It looks like new environment variables were added. We will have to add these to the ECS environment.

jeshuaborges (approved):
Excluding the one comment to add ECS environment variables, it looks good to me. @mmiermans can you take a quick pass and see if there are any improvements you would suggest?

src/utils/config.py
mmiermans: @gkatre I don't see `SNOWFLAKE_RAWDATA_DB` in [.aws/src/config/index.ts](https://github.com/Pocket/data-flows/pull/84/files#diff-562941ffdeea75262c91bd9b45ffaf2093301e296762d4a32e5feee4f1b35ba8) yet.

src/flows/data_retention/deleted_accounts_flow.py
```diff
@@ -1,46 +1,47 @@
...
-    (email)
-select distinct m.email
-from analytics.dbt_staging.stg_user_to_email_map as m
-join DELETED_USERS as u on u.user_id = m.user_id
-left join DELETED_EMAILS as e on e.email = m.email
-where e.email is null;
-"""
+
+@task()
+def query_file(file_name: str, **kwargs):
```
mmiermans, line 11:
```suggestion
```
**Optional:** This _does_ work, but it shouldn't be necessary to wrap Prefect tasks in a task. To be more consistent with how we run tasks elsewhere, you could instantiate `SnowflakeQueriesFromFile` in this module without using a function.

mmiermans (commented):
Nice job working with Prefect! I like that you've added queries to files. It's clear you're getting more familiar with this tool, and that's awesome to see. :muscle:

I think you might still need to add one environment variable to the .aws config?

src/flows/data_retention/deleted_accounts_flow.py
mmiermans: I now get the comment here, and this makes sense. Without using the function, we would've had to repeat setting the `account` and `user` parameters on every call.

PR #87, opened 2022-06-23
Add an automated test

mmiermans:
## Goal
- Allow us automatically run tests.
- Add a simple test that would've caught import errors.

Additionally, remove tests for pre-review and post-review:
- I believe they were failing for me locally.
- I don't consider them very readable.
- They will break if the implementation of the flows in question is changed.


jeshuaborges (approved):
Thank you

PR #89, opened 2022-06-24
Deleted User Account data deletion for Snapshot-Firehose raw data

gkatre:
## Goal

Adding the script for deleting raw data from the Snowflake Snapshot Firehose schema

## Implementation Decisions
The process framework for deleting data is similar to the previous tasks that execute on other schemas. In this case the tables have similar structure to the tables in the schema `RAW.FIREHOSE`
Here is the list of updates specific to this new task: 
- [x] Added environment variables: `SNOWFLAKE_SNAPSHOT_DB` and `SNOWFLAKE_SNAPSHOT_FIREHOSE_SCHEMA`
- [x] Updated `.aws/src/index.ts`, `README`
- [x] Added environment variables to `config.py`
- [x] Created Parameter store values: `/DataFlows/Prod/SNOWFLAKE_SNAPSHOT_FIREHOSE_SCHEMA` and `/DataFlows/Prod/SNOWFLAKE_SNAPSHOT_DB`


.aws/src/config/index.ts
```diff
@@ -44,13 +44,15 @@ const prefect = {
...
       'SNOWFLAKE_DATA_RETENTION_ROLE',
+      'SNOWFLAKE_DATA_RETENTION_WAREHOUSE',
       'SNOWFLAKE_DATA_RETENTION_DB',
       'SNOWFLAKE_DATA_RETENTION_SCHEMA',
       'SNOWFLAKE_SNOWPLOW_DB',
       'SNOWFLAKE_SNOWPLOW_SCHEMA',
       'SNOWFLAKE_RAWDATA_DB',
       'SNOWFLAKE_RAWDATA_FIREHOSE_SCHEMA',
-      'SNOWFLAKE_DATA_RETENTION_WAREHOUSE',
+      'SNOWFLAKE_SNAPSHOT_DB',
```
jeshuaborges, line 54: FWIW I am very intersted in making more sense of our databases and probably having fewer of them, meaning we would need fewer environment variables.

jeshuaborges (approved):
Looks great!

PR #91, opened 2022-06-28
Removing deleted accounts from Stripe

gkatre:
## Goal

- Adding script for removing Stripe deleted account data
- Adding a schedule to automate the deletion process to run on the 15th of every month

## Implementation Decisions
 
The criteria for identifying deleted user accounts is based on the `is_deleted=true` flag
Currently there are 2 tables (`CUSTOMER` and `CARD`) that have the  `is_deleted=true` flag set. The script deletes data from these 2 tables for the data identified as deleted.

## Reference
Slack conversation: https://pocket.slack.com/archives/C03E28D1GUD/p1654892275342369

### DIFF
--- a/src/flows/data_retention/deleted_accounts_flow.py
+++ b/src/flows/data_retention/deleted_accounts_flow.py
@@ -20,8 +20,8 @@ def query_file(file_name: str, **kwargs):
     )
 
 
-# with Flow(FLOW_NAME, schedule=get_cron_schedule(cron="0 0 1 * *")) as flow:
-with Flow(FLOW_NAME) as flow:
+with Flow(FLOW_NAME, schedule=get_cron_schedule(cron="0 0 15 * *")) as flow:
+
     # Maintain a list of deleted accounts in a protected DB/Schema tables
     add_deleted_users_result = query_file(file_name='deleted_account_users.sql',
                                           task_args=dict(name="SavingDeletedUserAccount_user_ids")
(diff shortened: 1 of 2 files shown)


kirill-demtchouk (approved):
Seems reasonable to me!

src/flows/data_retention/deleted_accounts_flow.py
```diff
@@ -20,8 +20,8 @@ def query_file(file_name: str, **kwargs):
     )
 
 
-# with Flow(FLOW_NAME, schedule=get_cron_schedule(cron="0 0 1 * *")) as flow:
-with Flow(FLOW_NAME) as flow:
+with Flow(FLOW_NAME, schedule=get_cron_schedule(cron="0 0 15 * *")) as flow:
```
jeshuaborges, line 23: FWIW we chose the 15th because it sounded nice and would run on a day when we are in the office for the first time. This would allow us to be present if there were any issues. G and I preferred this to a first automated run when everyone is on a long holiday weekend.

jeshuaborges (approved):
Looks great to me! Thank you!

src/flows/data_retention/deleted_accounts_flow.py
gkatre: Also, I have run this flow manually today and the deleted user account data has already been purged as of 9:30 AM pst

PR #92, opened 2022-07-07
Create flow to write editorial candidates to feature store

mmiermans:
## Goal
Load hardcoded candidates set from [a spreadsheet](https://docs.google.com/spreadsheets/d/1gdl9695Ib6Kd0OUX28yD03wJzRAcHiKZkZqn9DVdI2I/edit?usp=sharing) as a candidate set for Setup Moment onboarding.

Additional changes:
- I moved tasks shared between this flow and `setup_moment_editorial_candidates_flow.py` to `src/common_tasks/corpus_candidate_set.py`

## QA
1. Run the flow locally against Pocket-Dev.
2. Query Athena for corpus candidate sets:
  ```sql
  SELECT * FROM "sagemaker_featurestore"."development-corpus-candidate-sets-v1-1654266001" limit 10;
  ```

- [x] Expectation: It should include a record with id `57d544d6-0758-4cd1-a7b4-86f454c8eae8`.
- [x] Corpus items should contain `ID` and `TOPIC` keys.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-73

Spreadsheet:
* https://docs.google.com/spreadsheets/d/1gdl9695Ib6Kd0OUX28yD03wJzRAcHiKZkZqn9DVdI2I/edit?usp=sharing

Slack thread:
* https://pocket.slack.com/archives/C03EYPH93GU/p1656359792704609

### DIFF
--- a/src/common_tasks/corpus_candidate_set.py
+++ b/src/common_tasks/corpus_candidate_set.py
@@ -1,6 +1,6 @@
 import datetime
 import json
-from typing import Dict, Sequence
+from typing import Dict, Sequence, List
 
 import boto3
 from prefect import task, Parameter
@@ -9,6 +9,19 @@ from sagemaker.feature_store.feature_group import FeatureGroup
 from sagemaker.feature_store.inputs import FeatureValue
 
 from utils import config
+
+
+@task()
+def validate_corpus_items(corpus_items: List[Dict]):
+    min_item_count = 1
+    expected_keys = ['ID', 'TOPIC']
+
+    assert len(corpus_items) >= min_item_count
+    assert all(list(corpus_item.keys()) == expected_keys for corpus_item in corpus_items)
+    assert all(corpus_item['ID'] for corpus_item in corpus_items)
+    assert all(corpus_item['TOPIC'] for corpus_item in corpus_items)
+
+    return corpus_items
 
 
 @task()
--- a/src/common_tasks/corpus_candidate_set.py
+++ b/src/common_tasks/corpus_candidate_set.py
@@ -1,0 +1,37 @@
+import datetime
+import json
+from typing import Dict, Sequence
+
+import boto3
+from prefect import task, Parameter
+from sagemaker import Session
+from sagemaker.feature_store.feature_group import FeatureGroup
+from sagemaker.feature_store.inputs import FeatureValue
+
+from utils import config
+
+
+@task()
+def create_corpus_candidate_set_record(
+        id: str,
+        corpus_items: Dict,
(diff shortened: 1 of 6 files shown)


src/flows/recommendation_api/setup_moment_editorial_candidates_flow.py
```diff
@@ -0,0 +1,102 @@
...
+FROM "APPROVED_CORPUS_ITEMS"
+WHERE APPROVED_CORPUS_ITEM_EXTERNAL_ID IN ({",".join(f"'{corpus_id}'" for corpus_id in CORPUS_IDS)});
+"""
+
+
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=30)) as flow:
+    corpus_items = PocketSnowflakeQuery()(
+        query=EXPORT_CORPUS_ITEMS_SQL,
+        data={
+            'scheduled_at_start_day': -60,
```
jeshuaborges, line 87: Where are `scheduled_at_start_day` and `language` parameters used in the query?

src/flows/recommendation_api/setup_moment_editorial_candidates_flow.py
```diff
@@ -0,0 +1,102 @@
...
+            'language': 'EN',
+        },
+        database=config.SNOWFLAKE_ANALYTICS_DATABASE,
+        schema=config.SNOWFLAKE_ANALYTICS_DBT_SCHEMA,
+        output_type=OutputType.DICT,
+    )
+
+    feature_group_record = create_corpus_candidate_set_record(
```
jeshuaborges, line 95: How do you go from a query returning many records to a single feature group record being written? I would expect that there would be some map functions to process each record returned from the query.

src/common_tasks/corpus_candidate_set.py
```diff
@@ -0,0 +1,37 @@
...
+@task()
+def create_corpus_candidate_set_record(
+        id: str,
+        corpus_items: Dict,
+        unloaded_at: datetime.datetime = datetime.datetime.now()
+) -> Sequence[FeatureValue]:
+    return [
+        FeatureValue('id', id),
+        FeatureValue('unloaded_at', unloaded_at.strftime("%Y-%m-%dT%H:%M:%SZ")),
+        FeatureValue('corpus_items', json.dumps(corpus_items)),
```
jeshuaborges, line 23: Do we need to validate any properties in the corpus item dict to ensure that it matches the expectations of this record downstream? Additionally should we check to make sure that our size does not exceed 350KB?

https://docs.aws.amazon.com/sagemaker/latest/dg/feature-store-quotas.html

PR #93, opened 2022-07-08
Prospect training data flows

lfishhead:
## Goal
Goal here is to create flows that will update training data that is logged to feature groups.  The training data is used to retrain models that create prospecting candidate sets using global counts features and timespent features combined with publisher data and our content vectors stored in the vector store.

## Implementation Decisions

There are two flows: one for the global training data and the other for the timespent training data.  Each uses a `refresh` parameter that is set manually in the flow script to determine whether they are run incrementally (default) or to fully rebuild the training sets with backfill (pre-curation tool) data.  

The parameter governs which query is executed in snowflake to obtain data to be added to the feature group.  

There are separate feature groups for the two feature sets, both keyed on `resolved_id`.  We can change the schema later to key on a URL or corpus-api identifier.

### DIFF
--- a/src/flows/prospecting/global_training_features_flow.py
+++ b/src/flows/prospecting/global_training_features_flow.py
@@ -1,6 +1,7 @@
 import datetime
 import pandas as pd
 import boto3
+import re
 
 from prefect import task, Flow, Parameter, case
 from prefect.tasks.control_flow import merge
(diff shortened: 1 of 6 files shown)


lfishhead:
This is currently a draft while I validate its compatibility with the metaflow training job.  Any comments are welcome.

mmiermans (commented):
I focused on reviewing the Python code. Maybe @gkatre can review the SQL? There seems to be a lot of duplicate SQL code on first glance.

Overall the code looks good! All changes I suggested are minor, and it's clear you're getting better at working with Prefect. With Prefect 2.0, the code should get a lot more Pythonic instead of having to use custom Prefect concepts like `with case` and `map`.

gkatre:
@lfishhead we can perhaps review the SQLs during our paring time today?

PR #94, opened 2022-07-21
Add coronavirus candidates

eu9ene:
## Goal
Migrate generation of covid slate candidates from Metaflow to Prefect and from Elastic Search to Snowflake.

- Created Prefect flow
- Used Snowflake query provided by Matt
- Copied SQS writing code
- Added writing to SageMaker feature store like in Setup moment flow

## Implementation Decisions
 
- SQS code is temporary, so I put everything related in one file to remove it altogether.
- We can test it first and then merge it with a broader curated topic flow later.
- Removed writing to Firehose. Maybe we should add it later.

## QA
- Tested that it runs locally
- Deployed for testing in dev env - doesn't work, Prefect is broken
- Tested generated candidates with dev recapi 

## References

JIRA ticket:
* [Link to JIRA ticket](https://getpocket.atlassian.net/browse/HS-102)

Documentation:
* [Metaflow repo cleanup](https://getpocket.atlassian.net/wiki/spaces/~419497903/pages/2693431301/Metaflow+repo+cleanup)


src/api_clients/sqs.py
```diff
@@ -0,0 +1,114 @@
...
+        sqs_result = queue.send_message(MessageBody=message)
+
+        print(f"posting candidate set from {current.flow_name} with type {candidate_type} to firehose")
+        firehose = boto3.client('firehose')
+        # Firehose may batch records together, so they want newline delimited records
+        # See https://docs.aws.amazon.com/firehose/latest/APIReference/API_PutRecord.html
+        # So we are delivering new-line delimited UTF-8 encoded JSON
+        firehose_message = message + '\n'
+        firehose_result = firehose.put_record(DeliveryStreamName=firehose_name,
+                                              Record={"Data": firehose_message.encode('utf-8')})
```
mmiermans, line 89: I believe we use the Firehose to get data into Snowflake. Not sure why we chose to use a Firehose instead of Snowplow at the time. The pattern now is:
Application -> Snowplow -> Snowflake <-> Dbt modeling

Let's remove the Firehose code and keep the work out-of-scope for this PR for getting this data to Snowflake.

src/api_clients/sqs.py
```diff
@@ -0,0 +1,114 @@
...
+        return sqs_result, firehose_result
+    else:
+        print(f"put_results called from non-production {candidate_type} job; skipping put_results_sqs")
+
+
+# A Result set is a tuple, consisting of a guid, candidate list (recommended or curated prospects),
+# and a boolean indicating whether the candidate set is curated or not.
+ResultSetType = Tuple[str, Union[List[RecommendationCandidate], List[CurationProspect]], bool]
```
mmiermans, line 97: Nit: I would probably create a model for this type. (This might've been something that I wrote. :grimacing: )

src/flows/recommendation_api/coronavirus_flow.py
```diff
@@ -0,0 +1,71 @@
...
+            'language': 'EN',
+        },
+        database=config.SNOWFLAKE_ANALYTICS_DATABASE,
+        schema=config.SNOWFLAKE_ANALYTICS_DBT_SCHEMA,
+        output_type=OutputType.DICT,
+    )
+
+    #todo: blocklists?
+
+    # SafeMaker Feature Store won't be used until we switch in RecAPI
```
mmiermans, line 54:
```suggestion
    # SageMaker Feature Store won't be used until we switch in RecAPI
```

src/flows/recommendation_api/coronavirus_flow.py
```diff
@@ -0,0 +1,71 @@
...
+    print("Posting Coronavirus candidates to SageMaker Feature store.")
```
mmiermans, line 55: Any code in the `with Flow` block that is not a Prefect task will be run at build time instead of run-time, meaning that it will be executed in CodeBuild during deployment.

src/api_clients/sqs.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class NewTabFeedID(int, Enum):
+    en_US = 1
+    de_DE = 3
+    en_GB = 6
+    en_INTL = 8
+
+
+def put_results(candidate_set_id: str,
```
mmiermans, line 29: This needs an `@task` decorator for it to be called in a flow.

src/flows/recommendation_api/coronavirus_flow.py
```diff
@@ -0,0 +1,71 @@
...
+        data={
+            'scheduled_at_start_day': -60,
+            'language': 'EN',
+        },
+        database=config.SNOWFLAKE_ANALYTICS_DATABASE,
+        schema=config.SNOWFLAKE_ANALYTICS_DBT_SCHEMA,
+        output_type=OutputType.DICT,
+    )
+
+    #todo: blocklists?
```
mmiermans, line 52: Because content is now managed in the Corpus, we don't need a blocklist anymore! :raised_hands:

src/flows/recommendation_api/coronavirus_flow.py
```diff
@@ -0,0 +1,71 @@
...
+        corpus_items=corpus_items,
+    )
+    load_feature_record(feature_group_record, feature_group_name=feature_group)
+
+    print("Posting Coronavirus candidates to SQS.")
+    candidates = [RecommendationCandidate(
+        item_id=recommendation_item["resolved_id"],
+        publisher=recommendation_item["publisher"],
+        feed_id=NewTabFeedID.en_US
+    ) for recommendation_item in corpus_items]
```
mmiermans, line 67: Any transformations need to be done in a Prefect task, or they will be executed at build time instead of run-time.

mmiermans (commented):
Nice work! :100:

src/api_clients/sqs.py
mmiermans: ^ We can merge this PR in as long as we don't write to the production SQS queue, for example by removing the flow schedule.

src/api_clients/models.py
```diff
@@ -0,0 +1,41 @@
...
+    flow: str
+    run: str
+    expires_at: int
+
+    @root_validator(pre=False)
+    def validate_list_elements(cls, values):
+        expected_type = CurationProspect if values.get("type") == CandidateType.prospect else RecommendationCandidate
+        for c in values.get("candidates"):
+            assert type(c) == expected_type, f"expected {expected_type} : got {type(c)}"
+        return values
```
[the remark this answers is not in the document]
eu9ene: I think they're needed only for sqs for now which we're going to throw away soon so I wouldn't bother with those tiny refactoring. Also I really don't like applying word models to data classes :) It can cause confusion with ML models.

src/api_clients/sqs.py
```diff
@@ -0,0 +1,117 @@
...
+
+class CurationProspect(BaseModel):
+    scheduled_surface_guid: str
+    prospect_id: str
+    url: str
+    prospect_source: str
+    save_count: int
+    predicted_topic: str
+    rank: int
+
```
lfishhead, line 36: @eu9ene in case the context is missing, we attempted to reuse the SQS queue and code for two distinct purposes in the metaflow repo.  The first was generating the recommendation candidate sets for use in the recommendation-api.  The second was generating curation prospect candidate sets that are ingested by the curation tool (and editorial team).  that is how the two models above came to be.  we now have separate SQS queues as well

lfishhead:
this all looks great to me.

src/flows/recommendation_api/coronavirus_flow.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+FLOW_NAME = get_flow_name(__file__)
+
+CORONAVIRUS_CANDIDATE_SET_ID = 'c29fbf48-0093-49d2-ae55-4e01a3cf800e'
+
+CORONAVIRUS_SQL = """
+SELECT
+    s.resolved_id as id,   
+    c.top_domain_name as publisher,
```
mmiermans, line 30: TL;DR: For now, getting the publisher from `content.top_domain_name` is ok.

I see that the Corpus `publisher` field is not available in Snowplow or Snowflake at the moment. I've created [HS-116](https://getpocket.atlassian.net/browse/HS-116) to request this data.
- `JOIN content as c ON c.content_id = s.content_id` is not guaranteed to exist, because it relies  on an exact URL comparison between the Corpus API and a Parsed item.
- `top_domain_name` is not guaranteed to be accurate, because the Parser sometimes makes mistakes in extracting the domain from URLs (e.g. "co.uk" for bbc.co.uk), but for publisher spreading it is probably accurate enough.

mmiermans (approved):
Looks good :+1:

mmiermans (approved):

PR #98, opened 2022-08-02
adding candidate set generation for pocket hits slate

lfishhead:
## Goal

One new slate that is discussed for home shows items from pocket hits.  This is a simple slate to add now that doesn't exist in the recommendation API.

## Implementation Decisions
 
Basically followed @eu9ene work in the coronavirus flow quite closely.  The query for both the new flow and the coronavirus flow has been modified in the `WHERE` filter to limit items to be scheduled for the current date or earlier.  I realized that curators schedule items ahead of time in the new tool and we need this filter.


lfishhead:
@gkatre  and I reviewed this query together last week.

mmiermans (approved):

PR #99, opened 2022-08-02
Curated Corpus Topic Candidate sets to Feature store [HS-113]

gkatre:
## Goal

To generate recently Curated Corpus Topic Candidate item sets and ingest into Sagemaker Feature store

## Implementation Decisions

- `APPROVED_CORPUS_ITEMS` is queried for recent approvals (last 90 days) for corpus items
- The sets of Corpus topic item candidates are ingested to the `corpus-candidate-sets` Feature group
- Additional updates: Added/updated the test for corpus item IDs to be string (to prevent the old numeric IDs from being ingested)

JIRA ticket:
* [HS-113](https://getpocket.atlassian.net/browse/HS-113)

### DIFF
--- a/src/flows/recommendation_api/topics_flow.py
+++ b/src/flows/recommendation_api/topics_flow.py
@@ -1,5 +1,5 @@
 from prefect import Flow, Parameter, unmapped, task
-
+from prefect.executors import LocalDaskExecutor
 from api_clients.pocket_snowflake_query import PocketSnowflakeQuery, OutputType
 from common_tasks.corpus_candidate_set import (
     create_corpus_candidate_set_record,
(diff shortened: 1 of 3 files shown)


src/flows/recommendation_api/topics_flow.py
```diff
@@ -0,0 +1,59 @@
...
+FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS"
+WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
+AND TOPIC = %(CORPUS_TOPIC_ID)s
+AND LANGUAGE = 'EN'
+"""
+
+GET_TOPICS_SQL = """
+SELECT 
+    curated_corpus_candidate_set_id, 
+    corpus_topic_id
```
mmiermans, line 29: Optional suggestion: On line 35 we are relying on Snowflake turning these columns from lowercase to uppercase. We can remove this assumption by using double quotes to make our column names case-sensitive:
```suggestion
    "CURATED_CORPUS_CANDIDATE_SET_ID", 
    "CORPUS_TOPIC_ID"
```

mmiermans (approved):
Love the simplicity of this flow, especially compared with the code in Metaflow!

src/flows/recommendation_api/topics_flow.py
gkatre: I have updated the sql to alias the column names with the explicitly capitalized "CAP" field names.

Also, updated the return task object names to uniquely identify the tasks and their relationships in the flow

PR #100, opened 2022-08-04
Generate and ingest curated topic candidate sets (legacy) to SQS [HS-125]

gkatre:
## Goal

Generate and ingest curated topic candidate sets (legacy) to SQS
Benefits:
1. Consolidate the candidate set generation processes in a single flow using the Prefect service (reduces tech debit and code duplication in Metaflow)
2. Integrate Coronavirus (previously a special case) into a single flow with necessary metadata made available as a data model

## Deployment
- [ ] Carolyn has signed off on the changes in candidate sets ([Slack thread](https://pocket.slack.com/archives/C03EYPH93GU/p1659641383324719))

## Implementation Decisions

- Candidate set metadata previously imported and made available as data models are being used
- Curated Candidate set items are also available in the previously built data models making it possible to fetch the necessary data

### Observation:
The `STATIC_TOPIC_MAP_GUIDS` candidate set metadata previously imported from the static json file [topic_map_guids.json](https://github.com/Pocket/dl-metaflow-jobs/blob/main/jobs/resources/topic_map_guids.json#L28-L37) appears to have missing metadata `curatedCandidateID`, `algorithmicCandidateID` and `curatedNotNewTabCandidateID` for the `CORONAVIRUS` topic. For this PR we using the `curatedCandidateID` values.
We would need to update the seed data for the `CORONAVIRUS` topic for the missing values.
(Note: I have applied the [where CURATEDCANDIDATEID is not null and CORPUS_TOPIC_ID is not null](https://github.com/Pocket/data-flows/compare/legacy_topic_flow?expand=1#diff-749c58c9a3c210f4a04a17e86d299b8c58339ce62f6a9c26a1d41a35d1f3e556R32) filter to skip missing data)

## References

JIRA ticket:
* [HS-125](https://getpocket.atlassian.net/browse/HS-125)

### Test run in AWS Dev
The test run results using the AWS Dev account to send to the `RecommendationAPI-Dev` SQS and eventually to DynamoDB table `RECAPI-Dev-CandidateSets` is available [here](https://us-east-1.console.aws.amazon.com/dynamodbv2/home?region=us-east-1#item-explorer?initialTagKey=&table=RECAPI-Dev-CandidateSets)

### DIFF
--- a/src/common_tasks/corpus_candidate_set.py
+++ b/src/common_tasks/corpus_candidate_set.py
@@ -23,6 +23,18 @@ def validate_corpus_items(corpus_items: List[Dict]):
 
     return corpus_items
 
+@task()
+def validate_candidate_items(corpus_items: List[Dict]):
+    min_item_count = 1
+    expected_keys = ['ID', 'PUBLISHER']
+
+    assert len(corpus_items) >= min_item_count
+    assert all(list(corpus_item.keys()) == expected_keys for corpus_item in corpus_items)
+    assert all(isinstance(corpus_item['ID'], int) and corpus_item['ID'] != None for corpus_item in corpus_items)
+    assert all(isinstance(corpus_item['PUBLISHER'], str) and corpus_item['PUBLISHER'] != '' for corpus_item in corpus_items)
+
+    return corpus_items
+
 
 @task()
 def create_corpus_candidate_set_record(
--- a/src/flows/recommendation_api/legacy_topics_flow.py
+++ b/src/flows/recommendation_api/legacy_topics_flow.py
@@ -1,0 +1,70 @@
+from typing import List
+from prefect import Flow, Parameter, unmapped, task
+from prefect.executors import LocalDaskExecutor
+from api_clients.pocket_snowflake_query import PocketSnowflakeQuery, OutputType
+from common_tasks.corpus_candidate_set import (
+    validate_candidate_items,
+)
+from utils import config
+from utils.flow import get_flow_name, get_interval_schedule
+from api_clients.sqs import put_results, RecommendationCandidate, NewTabFeedID
+
+FLOW_NAME = get_flow_name(__file__)
+
+# Export approved candidate items by language and recency
+EXPORT_CORPUS_ITEMS_SQL = """
+SELECT 
+    a.resolved_id as "ID", 
+    c.top_domain_name as "PUBLISHER"
+FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" as a
+JOIN "ANALYTICS"."DBT".content as c ON c.content_id = a.content_id
+WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
+AND a.TOPIC = %(CORPUS_TOPIC_ID)s
+AND a.LANGUAGE = 'EN'
+"""
+
+GET_TOPICS_SQL = """
+select 
+    CURATEDCANDIDATEID as "CURATED_CORPUS_CANDIDATE_SET_ID",
+    CORPUS_TOPIC_ID as "CORPUS_TOPIC_ID"    
+from "ANALYTICS"."DBT"."STATIC_TOPIC_MAP_GUIDS" as m
+join "ANALYTICS"."DBT"."STATIC_TOPIC_LIST" as t on t.id = m.id
+where CURATEDCANDIDATEID is not null and CORPUS_TOPIC_ID is not null
+"""
+
+@task()
+def get_candidate_set_ids(topics):
+    return [i['CURATED_CORPUS_CANDIDATE_SET_ID'] for i in topics]
+
+@task()
+def transform_to_candidates(records: dict) -> List[RecommendationCandidate]:
+    return [RecommendationCandidate(
+        item_id=rec["ID"],
+        publisher=rec["PUBLISHER"],
+        feed_id=int(NewTabFeedID.en_US)
+    ) for rec in records]
+
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=30), executor=LocalDaskExecutor()) as flow:
+    query = PocketSnowflakeQuery(
+        database=config.SNOWFLAKE_ANALYTICS_DATABASE,
+        schema=config.SNOWFLAKE_ANALYTICS_DBT_SCHEMA,
+        output_type = OutputType.DICT
+    )
+
+    # Fetch a list of Topic Candidates
+    topics = query(query=GET_TOPICS_SQL)
+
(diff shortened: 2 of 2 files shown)


src/flows/recommendation_api/legacy_topics_flow.py
```diff
@@ -0,0 +1,70 @@
...
+    c.top_domain_name as "PUBLISHER"
+FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" as a
+JOIN "ANALYTICS"."DBT".content as c ON c.content_id = a.content_id
+WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
+AND a.TOPIC = %(CORPUS_TOPIC_ID)s
+AND a.LANGUAGE = 'EN'
+"""
+
+GET_TOPICS_SQL = """
+select
```
jeshuaborges, line 27: Im finding understanding what this query very difficult. Perhaps the table names should be more clear and we should add `corpus_topic_id` to the other table so we don't have to join.

mmiermans: Yeah, the name “topic map guids” is not very descriptive, unless you know there is a file in metaflow named like this. G, maybe you can create a static table with only the data that we need, and choose descriptive names without being ties to how things were named in Metaflow?

src/flows/recommendation_api/legacy_topics_flow.py
```diff
@@ -0,0 +1,70 @@
...
+FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" as a
+JOIN "ANALYTICS"."DBT".content as c ON c.content_id = a.content_id
+WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
+AND a.TOPIC = %(CORPUS_TOPIC_ID)s
+AND a.LANGUAGE = 'EN'
+"""
+
+GET_TOPICS_SQL = """
+select 
+    CURATEDCANDIDATEID as "CURATED_CORPUS_CANDIDATE_SET_ID",
```
mmiermans, line 28: This is not a corpus candidate set. Let’s call this column something like “legacy_curated_topic_candidate_set_id”?

src/flows/recommendation_api/legacy_topics_flow.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Fetch a list of Topic Candidates
+    topics = query(query=GET_TOPICS_SQL)
+
+    candidate_set_ids = get_candidate_set_ids(topics)
+
+    # Fetch the most recent Topic Candidate Items
+    topic_candidate_items = query.map(data=topics, query=unmapped(EXPORT_CORPUS_ITEMS_SQL))
+
+    # Fetch the most recent Topic Candidate Items
```
mmiermans, line 62: This comment applies to the previous line.

lfishhead:
The coronavirus page was distinct from the other topics.  In the old JSON this was indicated by the fact that it was an `editorial_collection` unlike the others.  There are no algorithmic items for it, and you ony need the one candidate set GUID.

My attitude is that most of the things that are unclear or clumsy here stem from decisions made in the creation of the recommendation-api.  It's hard to undo a lot of that, esp. when migrating candidate sets that are to be deprecated eventually?  I am curious why we aren't making the legacy slates pull corpus candidate sets, or how we anticipate doing that work later.

mmiermans:
> I am curious why we aren't making the legacy slates pull corpus candidate sets, or how we anticipate doing that work later.

The legacy slates are required to return a `resolved_id`. We could've decided to create hybrid candidate sets which identify both the CorpusItem and the `resolved_id`. Considering that we expect all legacy slates to be phased out, I think it's cleaner and easier to duplicate the flows and write to SQS, instead of having to change legacy slates to fetch candidates from the Feature Group.

src/flows/recommendation_api/legacy_topics_flow.py
gkatre: Yes. We can create a new static table in Snowflake dbt with better naming and just the 2 needed columns. We also need to include the missing metadata for the `CORONAVIRUS` topic in the table

src/flows/recommendation_api/legacy_topics_flow.py
gkatre: > This is not a corpus candidate set. Let’s call this column something like “legacy_curated_topic_candidate_set_id”?

We will have to do this in a separate Snowflake dbt PR. This is a Prefect PR that ingests the candidate set to SQS

mmiermans:
Let's add a deployment check box in the PR description to get sign-off from the editorial team:

- [ ] Carolyn has signed off on the changes in candidate sets ([Slack thread](https://pocket.slack.com/archives/C03EYPH93GU/p1659641383324719))

mmiermans:
Coronavirus candidate set id: c29fbf48-0093-49d2-ae55-4e01a3cf800e

lfishhead:
> Considering that we expect all legacy slates to be phased out.....
my question is when.  my concern is that we will maintain parallel versions of the same slates for a long time.

gkatre:
Updated the flow using the new static metadata table for legacy topic candidate set IDs

gkatre:
Based on the discussion with Matt, updated the query to limit the candidates per topic to 45.

@lfishhead , @mmiermans 
When you get a moment could both do a final review and approve for merge if things look ok?

src/common_tasks/corpus_candidate_set.py
```diff
@@ -23,6 +23,18 @@ def validate_corpus_items(corpus_items: List[Dict]):
 
     return corpus_items
 
+@task()
+def validate_candidate_items(corpus_items: List[Dict]):
+    min_item_count = 1
```
lfishhead, line 28: we may wish to increase this to 3 if we expect a full slate.

lfishhead (approved):
thanks G!

src/common_tasks/corpus_candidate_set.py
gkatre: Those are tests to verify the minimum state of the topic candidates and causing the flow to fail if those conditions are not met. 
If we are to increase the `min_item_count` to 3, we would have to apply filters to not have those topics when the candidates are less than 3.  Is that what you want?

mmiermans (approved):
:+1: lgtm

src/common_tasks/corpus_candidate_set.py
gkatre: @lfishhead I have merged the changes for now. If the candidates per topic min count 3 is important, lets discuss it and make the change in the next iteration

PR #101, opened 2022-08-10
Backfill topic preferences keyed on hashed_user_id [HS-132]

gkatre:
## Goal

To extend the User object with recommendation preferences we are creating a Feature group to key User preferences on the new  hashed user id. This Feature group will eventually the v1 based on the legacy data

## Implementation Decisions
 
Approach for backfilling user preferences feature group (v2) keyed on the hashed_user_id . The v1 feature group has the user preference data based on the numeric user_id.
The solution:

- pulls the data (Athena query - first time setup from Prefect)
- Snowflake query maps legacy user_id with hashed_user_id
- Perfect task merges the Athena + Snowflake datasets to build the dataframe for v2 feature group

## ToDos
- [ ] Create and setup the environment variable `ATHENA_S3_OUTPUT` in AWS Parameter store
- [ ] Assign a S3 location for `ATHENA_S3_OUTPUT`. This is the S3 staging location for Athena query execution results
- [ ] Necessary AWS permissions for the Athena query from Prefect to access read and write from the S3 location


src/api_clients/athena.py
```diff
@@ -0,0 +1,46 @@
...
+        sleep(1)
+
+    response = client.get_table_metadata(
+        CatalogName="AwsDataCatalog",
+        DatabaseName="sagemaker_featurestore",
+        TableName="development-user-recommendation-preferences-v1-1654826050",
+    )
+    columns = [
+        col['Name']
+        for col in response['TableMetadata']['Columns']
```
mmiermans, line 36: Use `response['ResultSet']['Rows']`

mmiermans:
Outside-in pattern we started during pair programming: https://github.com/Pocket/data-flows/pull/102

src/flows/recommendation_api/backfill_hashed_user_preferences.py
```diff
@@ -0,0 +1,80 @@
...
+    user_topics_prefs = AthenaQuery(
+        query=ATHENA_SQL,
+    )
+    user_topics_pref_rows = get_user_topics_pref_rows(user_topics_prefs)
+
+    # Prepare v2 user preferences using user_id to hash_user_id maps from Snowflake DB
+    v2_user_topics_prefs = build_v2_user_topics_prefs.map(v1_record=user_topics_pref_rows)
+
+    # Remove data with missing hash_user_id maps
+    v2_user_topics_prefs_clean = cleanup_v2_user_topics_prefs(v2_user_topics_prefs)
```
mmiermans, line 74: - Is there a reason you didn't follow a [fail-fast](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2754740314/Software+Design+Patterns#Fail-fast) approach here?
- Do you know that Snowflake will have missing records for the data we're interested in? Based on the [Slack thread](https://pocket.slack.com/archives/C02JZ4TRF0S/p1660068806342869) with Daniel, it was not clear that any records would be missing.
- This cleaning step makes me concerned that we might not be aware that the v2 store will have missing records if Snowflake does have missing data.

The 'fail-fast' approach would simply be not to include add this cleaning step:
```suggestion
```

src/api_clients/athena.py
```diff
@@ -0,0 +1,44 @@
+import boto3
+import pandas as pd
+from time import sleep
+from prefect import task
+from utils import config
+
+@task()
+def AthenaQuery(query: str = None):
```
mmiermans, line 8: - :snake: Python functions are styled using snake case
- `query` should probably be a required attribute
```suggestion
def athena_query(query: str):
```

src/flows/recommendation_api/backfill_hashed_user_preferences.py
```diff
@@ -0,0 +1,80 @@
...
+    hashed_user_id_map = query.run(
+        query=HASHED_USER_ID_MAP_QUERY,
+        data=v1_record,
+    )
+
+    return [
+        FeatureValue('hashed_user_id', hashed_user_id_map[0]['HASHED_USER_ID']),
+        FeatureValue('updated_at', v1_record['updated_at']),
+        FeatureValue('preferred_topics', v1_record['preferred_topics']),
+    ] if len(hashed_user_id_map) else None
```
mmiermans, line 59: The [fail-fast](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2754740314/Software+Design+Patterns#Fail-fast) approach would be:
```suggestion
    ]
```
The goal is:
- Spend less effort trying to be fault-tolerant
- By letting errors bubble up, errors and their root cause are easier to spot.

mmiermans (commented):
Some minor changes. I think you're almost there. Thanks for doing this backfill!

src/flows/recommendation_api/backfill_hashed_user_preferences.py
gkatre: > Do you know that Snowflake will have missing records for the data we're interested in? Based on the [Slack thread](https://pocket.slack.com/archives/C02JZ4TRF0S/p1660068806342869) with Daniel, it was not clear that any records would be missing

I have been running the flow in my local environment using data from the Dev Athena table `development-user-recommendation-preferences-v1-1654826050`. The data has `user_id` of `0` for a few records which will cause the failure if we remove the  `cleanup_v2_user_topics_prefs` function

gkatre:
@mmiermans Updated the PR based on your suggestions.

mmiermans (approved):
:+1: :ship:

PR #103, opened 2022-08-12
feat(publisher): move the publisher data job to prefect

bassrock:
## Goal

When talking to @jeshuaborges about moving a few remaining ETL style jobs in Web repo to AirFlow, he mentioned instead we should try moving them to Prefect.

This is the first take at that by moving this current hourly job https://github.com/Pocket/Web/blob/main/processes/publisher_data/update_pocket_score_data.py to Prefect. This is also the last Python job in the Web repo 🥳 

## Implementation Decisions

* The prefect docs have different run classes for MySQL query vs executing. This causes some weird duplication in code for the Pocket Publisher database class which is the same, except they extend a different class. 

## Deployment steps
- [x] Ask Data Products if they are ok with this move
- [x] Update the ENV variables with a pocket publisher database user https://github.com/Pocket/database-privileges/pull/133
- [x] Verify I coded this right
- [x] Verify I imported the right scipy version
- [x] Run locally against a dev version of the publisher database
- [ ] [Turn off the old cron script](https://github.com/Pocket/ansible-playbooks/blob/master/templates/etc/cron.d/Cron-app.crontab.j2#L6) https://github.com/Pocket/ansible-playbooks/pull/203


src/api_clients/pocket_publisher_database_execute.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(
+            self,
+            query: str = None,
+            database_host_env_var_name: str = "POCKET_PUBLISHER_DATABASE_HOST",
+            database_port_env_var_name: str = "POCKET_PUBLISHER_DATABASE_PORT",
+            database_dbname_env_var_name: str = "POCKET_PUBLISHER_DATABASE_DBNAME",
+            database_user_var_name: str = "POCKET_PUBLISHER_DATABASE_USER",
+            database_password_var_name: str = "POCKET_PUBLISHER_DATABASE_PASSWORD",
+            cursor_type: str = 'cursor',
```
mmiermans, line 26: [MySQLExecute](https://docs-v1.prefect.io/api/latest/tasks/mysql.html#mysqlexecute) doesn't have a `cursor_type` argument. Not sure if it would produce any errors, but probably better to remove all references to `cursor_type`.
```suggestion
```

src/api_clients/pocket_publisher_database_query.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(
+            self,
+            query: str = None,
+            database_host_env_var_name: str = "POCKET_PUBLISHER_DATABASE_HOST",
+            database_port_env_var_name: str = "POCKET_PUBLISHER_DATABASE_PORT",
+            database_dbname_env_var_name: str = "POCKET_PUBLISHER_DATABASE_DBNAME",
+            database_user_var_name: str = "POCKET_PUBLISHER_DATABASE_USER",
+            database_password_var_name: str = "POCKET_PUBLISHER_DATABASE_PASSWORD",
+            cursor_type: str = 'cursor',
```
mmiermans, line 26:
```suggestion
```
[MySQLFetch](https://docs-v1.prefect.io/api/latest/tasks/mysql.html#mysqlfetch) does have a `cursor_type` argument, and `'cursor'` is the default value, so all references to `cursor_type` can also be removed in this file.

src/api_clients/pocket_publisher_database_execute.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class PocketPublisherDatabaseExecute(MySQLExecute):
+    """
+    PocketPublisherDatabaseQuery automatically connects to our Publisher MySQL Database if the right environment variables are set:
+    - POCKET_PUBLISHER_DATABASE_HOST
+    - POCKET_PUBLISHER_DATABASE_PORT
+    - POCKET_PUBLISHER_DATABASE_DBNAME
+    - POCKET_PUBLISHER_DATABASE_USER
+    - POCKET_PUBLISHER_DATABASE_PASSWORD
```
mmiermans, line 15: If you add these variables to [parameterStoreNames](https://github.com/Pocket/data-flows/blob/main/.aws/src/config/index.ts#L35-L37) and create corresponding SSM Parameters, they will be injected in the environment.

src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -0,0 +1,196 @@
...
+    # The hack we are adding here creates a new content_quality_score table (_v2), fixes the column definitions
+    # and inserts new items into that table.
+    max_content_id = 2147483647
+
+    # score variables
+    loss_multiple = 9
+    inc_beta_var = float(1) / (1 + loss_multiple)
+    open_var = 1
+    fav_var = 5
+    share_var = 7
```
mmiermans, line 105: Nit: Code in the flow context is evaluated at build time. Let's move these constants out of the flow context, at the same scope as the above SQL string literals.

src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # prepare variables
+    first_hour_id = 1357020000
+    current_time = int(time.time())
+    today_hour_id = (current_time / 86400) * 86400
+    current_hour_id = today_hour_id - 86400
```
mmiermans, line 111:
```suggestion
```
- I don't see any references to these variables. Can we remove them?
- This code should be in a task to evaluate it at run-time, because otherwise it would be evaluated at build-time.
```python
@task()
def get_time_tuple():
    first_hour_id = 1357020000
    current_time = int(time.time())
    today_hour_id = (current_time / 86400) * 86400
    current_hour_id = today_hour_id - 86400
    return (first_hour_id, current_time, today_hour_id, current_hour_id)
```

src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -0,0 +1,196 @@
...
+
+    PocketPublisherDatabaseExecute()(
+        query=TRUNCATE_CONTENT_QUALITY_SQL
+    )
+
+    counter = 0
+
+    # loop through array of publisher_ids
+
+    while True:
```
mmiermans, line 125: Prefect 2.0 supports adding `while` and `for` loops to the flow context. We're on Prefect 1.x, so any loops would need to be moved inside a task. Tasks can then be triggered manually by calling `.run()` on the task:
```
@task()
def process_publisher_queue():
    while True:
        data = PocketPublisherDatabaseQuery().run(
            query=DATA_FOR_SCORE_CALCULATION_SQL
        )
        # Copy remaining code here, and call `.run()` on all tasks.
```
This is one of the features in 2.0 that makes me excited to upgrade.

mmiermans (commented):
Pretty amazing you got here, and I think it's almost done, too! :rocket:

mmiermans:
I [added dependencies between tasks](https://github.com/Pocket/data-flows/pull/103/commits/8e149c0e53032ee0b0d2462158f12bc95532f0bd). If you have graphviz installed and run `flow.visualize()`, Prefect produces the following graphic showing the dependencies:
Screenshot from 2022-08-17 14-50-14

mmiermans (approved):
lgtm :rocket:

bassrock:
>I am admittedly curious where the item score is used these days. Is this a stopgap before moving this data to snowflake or the feature store?

@lfishhead I do not know. I think in a few reports and in the elasticsearch item store?

PR #104, opened 2022-08-15
feat(Athena): Add permissions to query Athena [HS-154]

mmiermans:
## Goal
Add permissions to query Feature Stores using Athena.

## Implementation Decisions
1. Create the Athena query output bucket in Prefect, instead of using a shared bucket. This allows us to test in Pocket-Dev, without having to configure permissions and set an SSM Parameter for each environment.
2. Add Athena access to this policy, until [Pocket/data-shared](https://github.com/Pocket/data-shared) can be deployed to Pocket-Dev.

## QA
The [examples/athena_query](https://cloud.prefect.io/pocket/flow/6c3dbdf1-b3b7-4bdc-a6d3-26d3e43a8429) flow ran successfully:

Screenshot from 2022-08-15 10-42-51

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-154

Data-Shared permissions:
* https://github.com/Pocket/data-shared/blob/main/lib/permissions-stack.ts#L109-L161


gkatre (approved):
LGTM

PR #106, opened 2022-08-16
Update Athena get_query_results to iterate over entire result set

gkatre:
## Goal

Updates:
- Update Athena get_query_results to iterate over entire result set
- Logging missing user_id to hashed user_id maps in Snowflake DB


mmiermans (approved):

mmiermans:
We :pear:'ed on this.

PR #110, opened 2022-08-19
Publisher score data timeout

jeshuaborges:
## Goal

Make publisher data queue work the prefect way.

## Implementation Decisions
 
This updates publisher data flow to work how prefect intends into a ETL based job instead of what our old repo was doing. In particular, it will now re-use connections and execute in larger batches at more frequent intervals.

This was tested locally on my machine against the dev monolith database.


src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -32,53 +34,53 @@
...
+        fav_uv = %s, 
+        fav_dv = %s, 
+        share_uv = %s, 
+        share_dv = %s, 
+        open_score = round(%s,4), 
+        favorite_score = round(%s,4), 
+        share_score = round(%s,4), 
+        total_score = round(%s,4) 
+        where publisher_id = %s
+              and content_id = %s
```
mmiermans, line 83: Why was this changed from named to unnamed parameters?

src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -101,95 +103,98 @@
...
+    two_dict = [x for x in rows if x['content_id'] > max_content_id]
+
+    execute_many(query=INSERT_TO_CONTENT_V1_SQL, args=insert_args(one_dict))
+    execute_many(query=INSERT_TO_CONTENT_V2_SQL, args=insert_args(two_dict))
+
+    execute_many(query=UPDATE_CONTENT_V1_SQL, args=update_args(one_dict))
+    execute_many(query=UPDATE_CONTENT_V2_SQL, args=update_args(two_dict))
 
 
 with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=60)) as flow:
```
mmiermans, line 184: By default tasks run in a single thread. To leverage multi-threading with 16 threads:
```suggestion
with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=60), executor=LocalDaskExecutor(num_workers=16)) as flow:
```
`from prefect.executors import LocalDaskExecutor` at the top.

src/utils/config.py
```diff
@@ -40,6 +40,14 @@
     "schema": os.getenv('SNOWFLAKE_DATA_RETENTION_SCHEMA')
 }
 
+MYSQL_PUBLISHER_CONNECTION_DICT = {
+    "host": os.getenv("POCKET_PUBLISHER_DATABASE_HOST"),
+    "user": os.getenv("POCKET_PUBLISHER_DATABASE_USER"),
+    "db_name": os.getenv("POCKET_PUBLISHER_DATABASE_DBNAME"),
+    "port": int(os.getenv("POCKET_PUBLISHER_DATABASE_PORT")),
+    "password": os.getenv('POCKET_PUBLISHER_DATABASE_PASSWORD'),
+}
```
mmiermans, line 49: POCKET_PUBLISHER_DATABASE_PORT = os.getenv("POCKET_PUBLISHER_DATABASE_PORT")
```suggestion
MYSQL_PUBLISHER_CONNECTION_DICT = {
    "host": os.getenv("POCKET_PUBLISHER_DATABASE_HOST"),
    "user": os.getenv("POCKET_PUBLISHER_DATABASE_USER"),
    "db_name": os.getenv("POCKET_PUBLISHER_DATABASE_DBNAME"),
    "port": int(POCKET_PUBLISHER_DATABASE_PORT) if POCKET_PUBLISHER_DATABASE_PORT else None,
    "password": os.getenv('POCKET_PUBLISHER_DATABASE_PASSWORD'),
}
```

src/flows/publisher_data/update_pocket_score_data_flow.py
jeshuaborges: The pymysql execute multiple only accepts positional arguments. :(

mmiermans (approved):
:+1:

PR #114, opened 2022-08-23
Improve publisher data flow

jeshuaborges:
## Goal

Better error handling

* Retry loading data 3 times
* Change the schedule to every 20 minutes to ensure one job completes before the next execution
* Add logging statements for better visibility

https://pocket.slack.com/archives/C02KH5U7G79/p1661284371729469

### DIFF
--- a/src/flows/publisher_data/update_pocket_score_data_flow.py
+++ b/src/flows/publisher_data/update_pocket_score_data_flow.py
@@ -8,7 +8,7 @@ from prefect.tasks.mysql import MySQLExecute, MySQLFetch
 
 from api_clients.pocket_mysql import PocketMySQLExecuteMany
 from utils.config import MYSQL_PUBLISHER_CONNECTION_DICT
-from utils.flow import get_flow_name, get_interval_schedule
+from utils.flow import get_flow_name, get_interval_schedule, get_s3_result
 
 '''
 Load and update the content quality score in the publisher database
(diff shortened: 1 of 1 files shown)


src/flows/publisher_data/update_pocket_score_data_flow.py
```diff
@@ -173,29 +174,33 @@ def update_args(rows: [dict[str, Union[float, int]]]) -> [Tuple]:
...
+        inserts = execute_many.run(query=INSERT_TO_CONTENT_V2_SQL, args=insert_args(two_dict))
+        updates = execute_many.run(query=UPDATE_CONTENT_V2_SQL, args=update_args(two_dict))
+        deletes = execute_many.run(query=DELETE_PROCESS_QUEUE_SQL, args=delete_args(two_dict))
+        logging.info('V2 inserts: %s, updates: %s, deletes: %s', inserts, updates, deletes)
     else:
         logging.info('No publisher records to process for the V2 process')
 
 
-with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=10)) as flow:
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=20)) as flow:
```
mmiermans, line 203: I believe the retry will fail unless we specify result storage. Currently results are not stored by default.
```suggestion
with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=20), result=get_s3_result()) as flow:
```
Change line 11 to:
```python
from utils.flow import get_flow_name, get_s3_result
```

jeshuaborges: Great catch. Thank you.

bassrock (approved):
Makes sense to me!

mmiermans (approved):
🤞 Thanks for helping here! Hoping that our new hires won’t know what the Web Repo is.

PR #115, opened 2022-08-24
Publisher logging

jeshuaborges:
## Goal

Fix logging statements and add more retries


bassrock (approved):

mmiermans (approved):
Sorry that I didn’t catch this during review before.

PR #117, opened 2022-08-25
User impression flow updates

lfishhead:
## Goal
There were a bunch of recent failures with this flow so @mmiermans and I took another pass at it today.

## Implementation Decisions
 
The main change was to update the impression age measured to list items to be filtered.  Previously we were only updating the list if the item was impressed that day.  This is probably not the desired behavior since we were not concatenating the previous list with the new items.  In this PR, the attempt is to first look at all the impressions in the last 21 days.  These impression events are aggregated by (user, content_id) pair.  The impression age is computed as the difference between  the **current date** and the first time the user impressed the items instead of the most recent time the user impressed the item and the last time the user impressed the item.  

I am not sure this update to the feature group can be done incrementally as we had tried before for the reason that each day, all items the user has impressed increase in impression age, regardless of whether the user impressed them today.  

Small changes were made to change the feature group name to align with our more recent conventions.  I also changed from keying on the user ID to the hashed user ID which is also a new pattern.

### DIFF
--- a/src/flows/recommendation_api/user_impressions_flow.py
+++ b/src/flows/recommendation_api/user_impressions_flow.py
@@ -13,41 +13,35 @@ from utils.flow import get_flow_name
 FLOW_NAME = get_flow_name(__file__)
 
 BASE_QUERY = """
-WITH 
-prep as (
+WITH prep AS (
   SELECT
     a.HASHED_USER_ID,
     a.CONTENT_ID,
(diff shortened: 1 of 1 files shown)


gkatre:
@lfishhead I reviewed the query changes and they make sense to me. 👍

gkatre (approved):

PR #119, opened 2022-09-01
adding length based curated candidate generation flows

lfishhead:
## Goal

This adds two flows to prefect for candidate generation for recommendation-api candidate sets (which are actively being impressed).

* Curated Longreads: items from the recommended corpus with at least 3600 words ordered by review time (descending)
* Curate Shortreads:  items from the recommended corpus with at most 900 words ordered by review time (descending)

## Implementation Decisions

* for each, we've created candidate sets for English language and German language recommendations
* we're reusing the existing candidate set GUIDs for the English content
* limited candidate sets to 90 

## References

[google sheet](https://docs.google.com/spreadsheets/d/1uX6PJNegioGfQEyec4ybe4GVcetXs5xhWCigMNgA4_0/edit?usp=sharing) with slates that are getting impressions in August 2022

### DIFF
--- a/src/flows/recommendation_api/shortreads_flow.py
+++ b/src/flows/recommendation_api/shortreads_flow.py
@@ -13,7 +13,7 @@ from utils import config
 from utils.flow import get_flow_name, get_interval_schedule
 FLOW_NAME = get_flow_name(__file__)
 
-CURATED_SHORTREADS_CANDIDATE_SET_ID_EN = "626397d3-fa4e-4299-8d07-cbeaa269ebc1"
+CURATED_SHORTREADS_CANDIDATE_SET_ID_EN = "7ef90242-ff7a-44ac-8a32-53193e4a23eb"
 CURATED_SHORTREADS_CANDIDATE_SET_ID_DE = "57e4d3d1-9b4a-4a35-82f4-e577d88f6521"
 
 # Export approved corpus items by language and recency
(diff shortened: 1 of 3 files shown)


src/flows/recommendation_api/shortreads_flow.py
```diff
@@ -0,0 +1,73 @@
...
+# Export approved corpus items by language and recency
+EXPORT_LONGREADS_ITEMS_SQL = """
+SELECT 
+    a.resolved_id as "ID", 
+    c.top_domain_name as "PUBLISHER"
+FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" AS a
+JOIN "ANALYTICS"."DBT"."CONTENT" AS c
+  ON c.CONTENT_ID = a.CONTENT_ID
+WHERE a.REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD("day", -90, current_timestamp())
+AND c.WORD_COUNT <= 900
```
mmiermans, line 28: Nit: I wouldn't bother doing this for the legacy SQS flows, but when we create the Corpus flow, I'd probably make the `word_count` a parameter, and merge these flows into one.

mmiermans (approved):
lgtm

src/flows/recommendation_api/shortreads_flow.py
lfishhead: the tricky part about that is that the inequality has to flip also.  I considered this, but it probably would require two parameters and some kind of flag to turn things off.

PR #122, opened 2022-09-09
refactoring query for user impression capping data

lfishhead:
## Goal

This PR updates the user_impressions flow that aggregates items that users have impressed.  The change here revises the flow in light of our shift to corpus slates and corpus items in the recommendation-api.  The goal is to make the impression capping ranker part of the baseline implementation for web home.

## Implementation Decisions
 
The basic changes are to the query to snowflake.  To isolate impression events from corpus slates we join against the `STG_CORPUS_SLATE_RECOMMENDATIONS` model.  We use a feature group to store any items to be filtered that is keyed on the hashed user ID.  

The list contains any items that have been impressed more than 12 times by the user as well as any items that were first impressed at least 9 days ago.  The parameters can be adjusted here in prefect as needed.

Items are added to this list regardless of which slate or surface that the impression occurs.  All impression events originating with a recommendation-api request are aggregated.

### DIFF
--- a/src/flows/recommendation_api/user_impressions_flow.py
+++ b/src/flows/recommendation_api/user_impressions_flow.py
@@ -60,7 +60,7 @@ else:
     schedule = None
 
 with Flow(FLOW_NAME, schedule=schedule) as flow:
-    feature_group = Parameter("feature group", default=f"{config.ENVIRONMENT}-user-impressions-v1")
+    feature_group = Parameter("feature group", default=f"{config.ENVIRONMENT}-user-impressions-v2")
     max_impr_age = Parameter("max impression age", default=6)
     max_impr_count = Parameter("max impression count", default=9)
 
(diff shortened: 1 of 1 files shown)


mmiermans (approved):
Love it! :heart: So much simpler than the previous solution. By applying [outside-in development](https://en.wikipedia.org/wiki/Outside%E2%80%93in_software_development), we got exactly what we needed, and no more. Let's keep doing this going forward.

:ship:

src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -63,11 +62,11 @@ def load_feature_group(df: pd.DataFrame, feature_group_name):
 with Flow(FLOW_NAME, schedule=schedule) as flow:
     feature_group = Parameter("feature group", default=f"{config.ENVIRONMENT}-user-impressions-v1")
```
mmiermans, line 63:
```suggestion
    feature_group = Parameter("feature group", default=f"{config.ENVIRONMENT}-user-impressions-v2")
```
We probably need to rename the 'resolved_ids' column to 'corpus_ids'.

PR #123, opened 2022-09-09
Reorganize candidate selection and delete backfill flow

mmiermans:
## Goal
Reorganize flows under recommendation_api.

- Organize flows into corpus and legacy candidate set generation.
- Delete one-off backfill script for user topic preferences.

Screenshot from 2022-09-09 15-56-01

## Deployment steps
After deployment, I'll manually delete the old flows.


lfishhead (approved):
we may at some point want both `curated_topics` and `algorithmic_topics` but can cross that bridge when we get there.  as an aside, what tests do we need to do to figure out if we can remove the metaflow versions of longreads and shortreads?

PR #126, opened 2022-09-13
Scheduling fix

lfishhead:
## Goal

Small changes to the scheduling interval for the legacy candidate generation flows for the recommendation-api.

## Implementation Decisions
 
By conditioning on the environment variable that is set in config, we allow for running the flows immediately when testing in dev and running on a schedule when in production.

Changed interval for shortreads and longreads flows to 3 hours.


mmiermans (commented):
`get_interval_schedule` is a helper function that contains a condition on the environment, such that you don't have to copy that condition in every function.

PR #128, opened 2022-09-14
fix(legacy candiate sets): Short and long read schedules

mmiermans:
## Goal
- Enable shortreads schedule that was accidentally disabled.
- Changed interval for shortreads and longreads flows to 3 hours.


lfishhead (approved):
this is cleaner.  thanks much.

PR #130, opened 2022-09-15
feat(candidate set): Add newtab_en_us candidate set [HS-211]

mmiermans:
## Goal
Add candidate set for unpersonalized 'Recommended Reads' slate.
- Exclude syndicated per [content requirements](https://docs.google.com/document/d/15gBxnXlsKTgGyjZe72VUc43aK9ZoAArP9NaJDNLIaps/edit#heading=h.mljabq907ibf) 
- For now, rank based on recency. As a follow-up we can rank based on CTR, but this would require some modeling in Dbt.

## References

JIRA ticket:
* [Link to JIRA ticket](https://getpocket.atlassian.net/browse/HS-211)

Documentation:
- [Content requirements doc](https://docs.google.com/document/d/15gBxnXlsKTgGyjZe72VUc43aK9ZoAArP9NaJDNLIaps/edit#heading=h.mljabq907ibf)


lfishhead (approved):
LGTM, we may want to confirm if collections should be included or not

PR #131, opened 2022-09-19
Limit article text flow to 50 files

jeshuaborges:
## Goal

Limit article text flow to a number of files it can process in memory.

## Implementation Decisions

The flow hangs at ~77 files.
 
Screen Shot 2022-09-19 at 4 41 44 PM


gkatre (approved):
👍

PR #139, opened 2022-09-21
Update parser item processing to use more worker processes

jeshuaborges:
## Goal

Generate item text using all available ECS VCPU.

## Implementation Decisions

It appears that we are only using a single CPU core when converting HTML to text.

```
stats avg(MemoryUtilized), avg( ) by bin(5m)
| filter ispresent(ContainerName) and ContainerName like /flow/ and Type = "Container" and TaskId = '83578ffe78c64ecb93d5daac98b4d795'
| sort @timestamp desc
```

Screen Shot 2022-09-21 at 8 44 48 AM


gkatre (approved):
👍

PR #141, opened 2022-09-22
Hourly flow

jeshuaborges:
## Goal

Orchestrate all systems with dependencies to ensure that their upstream tasks are completed before executing downstream tasks. This should make sure that data is available more quickly even if it's executed less frequently.

## Implementation Decisions

1. There is probably a better way to make patterns of which flows to include. I started by being very explicit about what should be included and ensuring that we would get compiler errors quickly if there was a dependency misalignment. This will take more effort. As the patterns become more clear lets make more investments.
2. `hourly_flow.py` only triggers other flows and does not know anything about executing tasks directly.
3. I thought that it would make sense to treat folders with many flows as modules where they would have their own `hourly_flow.py` to ensure that this module tracked which flows it should include in the hourly process. This may be a good decision or bad. Im not sure, but its something that worked.
4. The dbt flows and downstream flows moved from 30min to hourly. I don't see anything in the `LOAD_FLOWS` which would have a problem with being triggered hourly instead of twice hourly.

### DIFF
--- a/src/flows/hourly_flow.py
+++ b/src/flows/hourly_flow.py
@@ -1,54 +1,44 @@
-from prefect import Flow, task, unmapped
-from prefect.tasks.dbt import dbt
-from prefect.tasks.prefect import create_flow_run, wait_for_flow_run
+from prefect import Flow, unmapped
 from prefect.executors import LocalDaskExecutor
-from utils import config
+from prefect.tasks.prefect import create_flow_run, wait_for_flow_run
 
+from flows import dbt_hourly_flow, parser_item_html_text_streaming_flow
 from flows.braze import update_flow as braze_update_flow
 from flows.prospecting import prereview_engagement_feature_store_flow, postreview_engagement_feature_store_flow
-from flows import dbt_hourly_flow
+from flows.recommendation_api.corpus_candidate_sets import hourly_flow as corpus_candidate_sets_hourly_flow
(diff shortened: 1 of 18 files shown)


src/flows/hourly_flow.py
```diff
@@ -0,0 +1,44 @@
...
+
+LOAD_FLOWS = [
+    braze_update_flow.FLOW_NAME,
+    prereview_engagement_feature_store_flow.FLOW_NAME,
+    postreview_engagement_feature_store_flow.FLOW_NAME,
+    corpus_candidate_sets_hourly_flow.FLOW_NAME,
+    legacy_candidate_sets_hourly_flow.FLOW_NAME,
+]
+
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=60), executor=LocalDaskExecutor()) as flow:
```
mmiermans, line 31: The default `num_workers` is equal to the number of cores, which would prevent all flows from running in parallel if we have more than 4 flows on a 4 vCPU instance.
```suggestion
with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=60), executor=LocalDaskExecutor(scheduler="threads", num_workers=32)) as flow:
```

mmiermans (commented):
> This should make sure that data is available more quickly

Was there a particular use-case in which the data wasn't available fast enough?

I don't love the idea of adding a dependency on the text streaming flow to load flows that don't use this data. Should we wait until we have a better strategy, or is there a particular use-case that drives this change?

PR #142, opened 2022-09-23
Flow to split large files before Snowflake import

gkatre:
## Goal

To split the large Aurora Export files into smaller manageable files before import into Snowflake 

## Implementation Decisions
 - Flow `parser_item_html_backfill_filesplit_flow`: Splits the files into smaller files
 - After the  `parser_item_html_backfill_filesplit_flow` completes the `parser_item_html_text_backfill_flow` Flow-of-Flows is launched that invokes a map of  `parser_item_html_text_backfill_file_flow` flows that imports the Article content into Snowflake


src/flows/parser_item_html_backfill_filesplit_flow.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@task()
+def get_source_keys() -> [str]:
+    """
+    :return: List of S3 keys for the S3_BUCKET and SOURCE_PREFIX
+    """
+    logger = prefect.context.get("logger")
+
+    keys = S3List().run(bucket=SOURCE_S3_BUCKET, prefix=SOURCE_PREFIX)
```
jeshuaborges, line 37: You later in this flow you are passing dataframes between tasks. This will load all data matching this prefix into python memory. You can only process as much data in a flow that you can fit it all into memory.

src/flows/parser_item_html_backfill_filesplit_flow.py
```diff
@@ -0,0 +1,98 @@
...
+        raise Exception(f'No files to process for s3://{SOURCE_S3_BUCKET}/{SOURCE_PREFIX}.')
+
+    if len(keys) > NUM_FILES_PER_RUN:
+        logger.warn(f"Number of files is greater than the number a worker can process in a single run. Found {len(keys)} files, processing {NUM_FILES_PER_RUN}.")
+        return keys[0:NUM_FILES_PER_RUN]
+    else:
+        return keys
+
+
+@task(timeout=10 * 60, max_retries=18, retry_delay=timedelta(seconds=10))
```
jeshuaborges, line 48: I suggest you remove retries. If you don't remove them you will have to stage each task result to s3.

jeshuaborges (changes requested):
Retries wont work and you wont be able to process more than one file.

src/flows/parser_item_html_backfill_filesplit_flow.py
gkatre: @jeshuaborges good point. 
I was thinking about reducing the count on `NUM_FILES_PER_RUN`.
but then, the main purpose of this flow is to only split the files into smaller sized and not take excessive time for processing.
What are your thoughts about not using mappers at all for this flow?

jeshuaborges (approved):
Looks great

PR #143, opened 2022-09-23
Add a text md5 field to article content table

jeshuaborges:
## Goal

Reduce the compute cost for down stream processes needing to dedupe articles.

@gkatre can you help me update the production table to accept this new field and backfill records with a null `text_md5` value?


gkatre:
@jeshuaborges added the new field for both streaming and backfill tables:
```sql
alter table raw.item.article_content_v2 add column text_md5 string; 
alter table snapshot.item.article_content_v2 add column text_md5 string; 
```

gkatre (approved):

PR #156, opened 2022-10-04
filtering topic corpus candidate sets

lfishhead:
## Goal
In preparation for unified home QA we are changing the curated topic candidate sets to filter items that are recommended during curation for the `NEW_TAB_EN_US` scheduled surface.

## Implementation Decisions
 
Changed topic query that previously filtered only on `LANGUAGE = 'EN'` to filter instead on `SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'` which implies english content.

This got revised a bit.  To add manual submissions, we need to include items from `scheduled_corpus_items` also, which is now included.  Some items in that subquery were showing up more than once which surprised me since they should not have been able to run more than once on en-US new tab.  These may be syndicated or collections so we need some guidance here on content filtering.

Tested the flow in the dev account.


jeshuaborges (approved):
Dang, look at that data model go.

src/flows/recommendation_api/corpus_candidate_sets/topics.py
```diff
@@ -14,13 +14,38 @@
...
+    AND SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
+    GROUP BY 1, 2
+    
+    )
+
+SELECT
+    ID,
+    TOPIC
+FROM PREP
+ORDER BY REVIEW_TIME DESC
```
mmiermans, line 48:
```suggestion
QUALIFY row_number() OVER (PARTITION BY ID ORDER BY REVIEW_TIME DESC) = 1
```

src/flows/recommendation_api/corpus_candidate_sets/topics.py
```diff
@@ -14,13 +14,38 @@
...
+    SELECT 
+        approved_corpus_item_external_id as "ID", 
+        topic as "TOPIC",
+        MAX(reviewed_corpus_item_updated_at) as "REVIEW_TIME" 
+    FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" 
+    WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
+    AND CORPUS_REVIEW_STATUS = 'recommendation'
+    AND TOPIC = %(CORPUS_TOPIC_ID)s
+    AND SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
+    GROUP BY 1, 2
```
mmiermans, line 27: Remove `max` and `group by`? Same suggestion for other query in the union.
- We should use the latest topic value, in case curators change them.
- There's no need to deduplicate here if we do it at the end.
```suggestion
    SELECT 
        approved_corpus_item_external_id as "ID", 
        topic as "TOPIC",
        reviewed_corpus_item_updated_at as "REVIEW_TIME" 
    FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" 
    WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', -90, current_timestamp())
    AND CORPUS_REVIEW_STATUS = 'recommendation'
    AND TOPIC = %(CORPUS_TOPIC_ID)s
    AND SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
```

mmiermans:
We'll probably want to apply a similar change to [recommended_by_recency.py](https://github.com/Pocket/data-flows/blob/main/src/flows/recommendation_api/corpus_candidate_sets/recommended_by_recency.py#L18-L26). That can be a follow-up PR.

src/flows/recommendation_api/corpus_candidate_sets/topics.py
```diff
@@ -14,13 +14,38 @@
...
+    
+    UNION
+    
+    SELECT 
+        approved_corpus_item_external_id as "ID", 
+        topic as "TOPIC",
+        MAX(scheduled_corpus_item_scheduled_at) as "REVIEW_TIME"
+    FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS"
+    WHERE SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD('day', -90, current_timestamp()) AND current_timestamp()
+    AND CORPUS_ITEM_LOADED_FROM = 'MANUAL'
```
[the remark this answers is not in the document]
lfishhead: no.  this filters for manually submitted items.

mmiermans (approved):
:+1: Thanks for taking this on!

PR #166, opened 2022-10-05
Try to fix bug where multiple flow runs are not created

jeshuaborges:
## Goal

Force a new flow run for each chunk.

Not creating a name reused the same flow run for each chunk. This is obviously not what we want to do.

[logs.txt](https://github.com/Pocket/data-flows/files/9719521/logs.txt)

### DIFF
--- a/src/flows/parser_item_html_backfill_load_all_flow.py
+++ b/src/flows/parser_item_html_backfill_load_all_flow.py
@@ -1,4 +1,4 @@
-import math
+from datetime import datetime
 
 import numpy as np
 import prefect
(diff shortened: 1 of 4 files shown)


src/flows/parser_item_html_backfill_load_all_flow.py
```diff
@@ -39,10 +39,17 @@ def get_source_keys(num_files: int) -> [str]:
...
-    for keys in np.array_split(keys, num_workers):
-        logger.info(f"Queuing worker for keys: {*keys,}.")
+    idx = 0
+    name_prefix = f'{FLOW_NAME}-{datetime.now().strftime("%m%d%Y-%H:%M:%S")}'
+    chunks = np.array_split(keys, num_workers)
+    for chunk in chunks:
+        logger.info(f"Queuing worker for keys: {*chunk,}.")
+        # testing adding run_name to the parameters because without it prefect is kicking off the same flow run many
+        # times with different parameters and only the last one is succeeding. The hope is that with a name it will
+        # create a new flow run.
```
mmiermans, line 49: :thinking: That's unexpected.. Over half of our Prefect PRs would've been easier under Prefect 2.0. Glad we're tracking to upgrade relatively soon! :rocket:

mmiermans (approved):

PR #175, opened 2022-10-10
fix(Candidates): Collections should be ranked by creation & scheduled date [HS-288]

mmiermans:
## Goal
Rank Collection candidates in reverse chronological order by the Corpus Item creation date or by the Corpus Item schedule date on New Tab en-US, whichever is most recent.

This is a workaround to mimic the Collection publish date, which is not available in Snowflake.

## Implementation Decisions
- Exclude scheduled dates that are in the future.

## Deployment steps
- [x] Flow runs successfully in Pocket-Dev

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-288

### DIFF
--- a/src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
+++ b/src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
@@ -15,15 +15,30 @@ FLOW_NAME = get_flow_name(__file__)
 RECENT_COLLECTIONS_CANDIDATE_SET_ID = "92af3dae-25c9-46c3-bf05-18082aacc7e1"
 
 EXPORT_COLLECTIONS_CANDIDATE_SET_SQL = """
-SELECT 
+WITH recent_collections AS (
+  SELECT
+    IFNULL(s.SCHEDULED_CORPUS_ITEM_SCHEDULED_AT, a.REVIEWED_CORPUS_ITEM_CREATED_AT) as recency,
+    s.SCHEDULED_SURFACE_IANA_TIMEZONE,
+    a.*
+  FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" a
+  LEFT JOIN "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS" s ON (
+    s.approved_corpus_item_external_id = a.approved_corpus_item_external_id
+    AND s.SCHEDULED_CORPUS_ITEM_SCHEDULED_AT < current_timestamp()
+    AND s.SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
+  )
+  WHERE a.IS_COLLECTION
+  AND a.CORPUS_REVIEW_STATUS = 'recommendation'
+  AND a.LANGUAGE = 'EN'
+  AND recency > DATEADD("day", %(MAX_AGE_DAYS)s, current_timestamp())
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
```diff
@@ -15,15 +15,30 @@
...
+WITH recent_collections AS (
+  SELECT
+    IFNULL(s.SCHEDULED_CORPUS_ITEM_SCHEDULED_AT, a.REVIEWED_CORPUS_ITEM_CREATED_AT) as recency,
+    s.SCHEDULED_SURFACE_IANA_TIMEZONE,
+    a.*
+  FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" a
+  LEFT JOIN "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS" s ON (
+    s.approved_corpus_item_external_id = a.approved_corpus_item_external_id
+    AND s.SCHEDULED_CORPUS_ITEM_SCHEDULED_AT < current_timestamp()
+    AND s.SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
```
lfishhead, line 27: I am curious if the collections that are not scheduled will be associated with any scheduled surface.  I am also curious how collections get into the prospecting tool or the `approved_corpus_items` model at all.  They are not part of any prospect source, so I am worried they will not have an associated surface (which may be fine).

src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
```diff
@@ -15,15 +15,30 @@
...
+  )
+  WHERE a.IS_COLLECTION
+  AND a.CORPUS_REVIEW_STATUS = 'recommendation'
+  AND a.LANGUAGE = 'EN'
+  AND recency > DATEADD("day", %(MAX_AGE_DAYS)s, current_timestamp())
```
lfishhead, line 32: this should probably be something like `BETWEEN DATEADD("day", %(MAX_AGE_DAYS)s, current_timestamp()) AND current_timestamp()` to avoid showing items that are scheduled in the future as is a common practice

lfishhead (approved):
approving tho it's likely changes are needed.  i would discuss with Carolyn if showing items as scheduled is what's expected.  if so, you can probably simplify the query by restricting it to the `scheduled_corpus_items` model.  if not, we need to see how collections get into the `approved_corpus_items model` if they are not scheduled, or if that even happens.  i.e. is there an expectation to show newly approved collections that aren't running in new tab.

lfishhead:
This query returns no results for me:
```
SELECT
    a.*
FROM ANALYTICS.DBT.APPROVED_CORPUS_ITEMS AS a
JOIN ANALYTICS.DBT.SCHEDULED_CORPUS_ITEMS AS s
  ON a.approved_corpus_item_external_id = s.approved_corpus_item_external_id
WHERE a.IS_COLLECTION
  AND s.scheduled_corpus_item_external_id IS NULL
  ```

src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
mmiermans: Those items are removed by the condition on line 26.

src/flows/recommendation_api/corpus_candidate_sets/collections_by_recency.py
mmiermans: Technically curators could submit Collections manually and recommend them, without being scheduled on New Tab. The last time that happened was in Dec 2021 for a 'most read articles' collection.

PR #177, opened 2022-10-11
fix(Candidates): recommended_by_recency always has at least 6 items p…

mmiermans:
## Goal
There's a new product requirement that if one item in a slate has a recommendation reason (topic bubble), then all items in that slate should have one.

This was not the case when we had insufficient recommendations for the topics that the user preferred. This change selects the 6 most recent items for each topic, which implicitly causes all items in 'For You' (a 3x2 slate) to have a recommendation reason.

## Implementation Decisions
- I split up the query in a few more cases to make clearer what each subquery does.
- I used `union all` instead of `union`. The former is more performant as our corpus grows. It has no effect on the query result because we already deduplicate using `qualify`.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-304

## QA
- [x] Run the query in Snowflake (see screenshot)
   - [x] Each topic has 6 candidates
   - [x] Review dates are recent
- [x] Run the flow against Pocket-Dev

Screenshot from 2022-10-11 13-25-26

### DIFF
--- a/src/flows/recommendation_api/corpus_candidate_sets/recommended_by_recency.py
+++ b/src/flows/recommendation_api/corpus_candidate_sets/recommended_by_recency.py
@@ -16,45 +16,55 @@ RECENT_RECOMMENDED_CANDIDATE_SET_ID = "2066c835-a940-45ec-b1f7-267457d9e0a2"
 
 
 EXPORT_RECOMMENDED_CANDIDATE_SET_SQL = """
-WITH prep as (
+WITH recently_updated_items as (
     SELECT 
         approved_corpus_item_external_id as "ID", 
         topic as "TOPIC",
         reviewed_corpus_item_updated_at as "REVIEW_TIME" 
     FROM "ANALYTICS"."DBT"."APPROVED_CORPUS_ITEMS" 
-    WHERE REVIEWED_CORPUS_ITEM_UPDATED_AT >= DATEADD('day', %(MAX_AGE_DAYS)s, current_timestamp())
-    AND CORPUS_REVIEW_STATUS = 'recommendation'
+    WHERE CORPUS_REVIEW_STATUS = 'recommendation'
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/corpus_candidate_sets/recommended_by_recency.py
```diff
@@ -16,45 +16,55 @@
...
+    WHERE SCHEDULED_CORPUS_ITEM_SCHEDULED_AT < current_timestamp()
     AND CORPUS_ITEM_LOADED_FROM = 'MANUAL'  -- should this be removed?
     AND SCHEDULED_SURFACE_ID = 'NEW_TAB_EN_US'
     AND NOT is_syndicated
     AND NOT is_collection
-    
-    )
+),
+
+all_recent_items as (
```
lfishhead, line 44: I'd probably combine these two CTEs (which has no functional difference), curious if G sees any performance difference
```
deduped as (
  SELECT * FROM (
    SELECT * FROM recently_scheduled_items
    UNION ALL 
    SELECT * FROM recently_updated_items )
  QUALIFY row_number() OVER (PARTITION BY ID ORDER BY REVIEW_TIME DESC) = 1
)
```

gkatre: There is no performance difference. the `all_recent_items` is just an alias for the same UNION ALL sql you are doing. Mathijs may have done it for readability reasons

gkatre (approved):
@mmiermans LGTM! 👍

mmiermans:
@gkatre That's right. I actually based it on something you did here: https://github.com/Pocket/dbt-snowflake/blob/main/models/staging/recommendation/stg_corpus_slate_recommendations.sql#L90-L94

PR #179, opened 2022-10-17
Updating lifehacks 

lfishhead:
## Goal

I believe the slates that are needed for unified home v1 are pocket hits and lifehacks.  This PR just updates the lifehacks slate with the same fix we applied in #156 

## Implementation Decisions

again we are getting manual submissions from the scheduled corpus items table and approved prospects from the approved corpus items table.  

## References

Documentation:
* [figma](https://www.figma.com/file/jawk2lhpNJZZWSJqE3DuNn/Unified-Home-Experiments?node-id=153%3A12544)


mmiermans (approved):
:+1: Thank you!

PR #180, opened 2022-10-25
Modify Pocket Hits candidates

mmiermans:
## Goal
Limit Pocket Hits candidate set to today's Pocket Hits email if available. Only backfill with older content if not enough content is scheduled for today.

## Implementation decision
- Assume email contains 8 stories. I think the emails currently always contain 8 stories, and I couldn't immediately think of a query to dynamically pull all stories from today, or backfill them with older stories until we have at least 8 of them.

## References

https://getpocket.atlassian.net/browse/HS-163

### DIFF
--- a/src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
+++ b/src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
@@ -21,13 +21,14 @@ POCKETHITS_EN_CANDIDATE_SET_ID = "92411893-ebdb-4a43-ad29-aa79e56e2136"
 
 POCKETHITS_SQL = """
 SELECT
-    APPROVED_CORPUS_ITEM_EXTERNAL_ID as ID,
-    TOPIC
-FROM "SCHEDULED_CORPUS_ITEMS"
+    APPROVED_CORPUS_ITEM_EXTERNAL_ID as "ID",
+    TOPIC as "TOPIC"
+FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS"
 WHERE SCHEDULED_SURFACE_ID = %(SURFACE_GUID)s
 AND SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD(day, %(MAX_AGE_DAYS)s, CURRENT_DATE) AND CURRENT_DATE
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
```diff
@@ -21,13 +21,14 @@
 
 POCKETHITS_SQL = """
 SELECT
-    APPROVED_CORPUS_ITEM_EXTERNAL_ID as ID,
-    TOPIC
-FROM "SCHEDULED_CORPUS_ITEMS"
+    APPROVED_CORPUS_ITEM_EXTERNAL_ID as "ID",
+    TOPIC as "TOPIC"
+FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS"
```
lfishhead, line 26: doesn't have to be here, but for publisher spread, eventually all these queries will need publisher fields

src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
```diff
@@ -21,13 +21,14 @@
...
 WHERE SCHEDULED_SURFACE_ID = %(SURFACE_GUID)s
 AND SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD(day, %(MAX_AGE_DAYS)s, CURRENT_DATE) AND CURRENT_DATE
 QUALIFY row_number() OVER (PARTITION BY APPROVED_CORPUS_ITEM_EXTERNAL_ID ORDER BY SCHEDULED_CORPUS_ITEM_SCHEDULED_AT DESC) = 1
 ORDER BY SCHEDULED_CORPUS_ITEM_SCHEDULED_AT DESC
+LIMIT 8  -- Only include past Pocket Hits stories if today's aren't available. There are 8 stories per email.
```
lfishhead, line 31: again, we can discuss, but my view is a top8 ranker would be better.  this would allow a larger candidate set that could be reused for personalization over a set from a larger time window.  this approach is constraining the candidate set with the slate requirements.

lfishhead (approved):

src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
mmiermans: I view the candidate set as all items that are eligible to be shown. The goal of this candidate set is to have timely Pocket Hits stories. The `limit 8` achieves that because each day 8 Pocket Hits stories are scheduled. I agree that it would be better to not make that assumption here, and instead filter only on `scheduled_at`.

PR #181, opened 2022-10-27
adding publisher field to corpus slates

lfishhead:
## Goal

We are adding a ranker for corpus slates that spreads items according to both their publisher and their topic.  To enable this ranker, the corpus slates must include a publisher field.

## Implementation Decisions
 
We made the publisher field optional for the time being, allowing NULL values which we encounter primarily among syndicated items and collections.  

## References

[HS-281](https://getpocket.atlassian.net/browse/HS-281)


mmiermans (approved):
Thanks for moving this forward. :bow:

PR #183, opened 2022-10-31
fix(Pocket Hits candidates): Override scheduled time to be 10am EDT [HS-358]

mmiermans:
## Goal
Start showing today's Pocket Hits content at 3am EST instead of 12am UTC.

Additional change: Remove condition that content can be at most 3 days old. This had no effect because of the sort order and `LIMIT 8` of most recent stories.

## Implementation Decisions
- For SQL efficiency it would be better not to manipulate a timestamp before doing a comparison, but I couldn't think of a way to meet the requirement without doing this.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-358

Documentation:
* [Content v1 slates spec](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2820046849/Unified+Home+V.1+Enhanced+Experimentation+Slates)

### DIFF
--- a/src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
+++ b/src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
@@ -26,7 +26,9 @@ SELECT
     PUBLISHER as "PUBLISHER"
 FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS"
 WHERE SCHEDULED_SURFACE_ID = %(SURFACE_GUID)s
-AND SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD(day, %(MAX_AGE_DAYS)s, CURRENT_DATE) AND CURRENT_DATE
+-- BACK-1668: `scheduled_at` is always 12am UTC, which is unexpected because Pocket Hits goes out at 10am EST.
+--            The workaround below overrides the time to be 2pm UTC = 10am EST. Can be removed when BACK-1668 is fixed.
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/corpus_candidate_sets/pockethits.py
```diff
@@ -26,7 +26,9 @@
     PUBLISHER as "PUBLISHER"
 FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS"
 WHERE SCHEDULED_SURFACE_ID = %(SURFACE_GUID)s
-AND SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD(day, %(MAX_AGE_DAYS)s, CURRENT_DATE) AND CURRENT_DATE
+-- Pocket Hits content should be available on Home at 3am EST = 7am UTC. Override the scheduled time to 7am.
+-- BACK-1668: `scheduled_at` is always 12am UTC, which is unexpected because Pocket Hits goes out at 10am EST.
+AND DATEADD('hour', 7, DATE_TRUNC('DAY', SCHEDULED_CORPUS_ITEM_SCHEDULED_AT)) < CURRENT_TIMESTAMP
```
lfishhead, line 31: i don't know how to make this better either, but we should follow up with G to see if he knows anything

lfishhead (approved):
I was a little confused on the time math, but trust you thouht about it.  scheudled_at is currently 12 UTC and you want to make it 10 EST (in UTC) with the -7 I assume?

mmiermans:
> I was a little confused on the time math, but trust you thouht about it.

I tried to clarify the comments a bit to explain how I'm setting the time to 7am UTC = 3am EST.

PR #184, opened 2022-11-07
data validation on user_impressions

lfishhead:
## Goal

user_impressions flow was [failing](https://pocket.slack.com/archives/C02KH5U7G79/p1667779458650989) on a data validation step that reported that `HASHED_USER_ID` was missing.

## Implementation Decisions

Added a line to drop rows that are missing this field.  I find it hard to figure out how many rows are being loaded into the feature group or log it to the console.

### DIFF
--- a/src/flows/recommendation_api/user_impressions_flow.py
+++ b/src/flows/recommendation_api/user_impressions_flow.py
@@ -27,9 +27,9 @@ WITH prep AS (
 )
 
 SELECT 
-    HASHED_USER_ID,
-    current_date as UPDATED_AT,
-    ('[' || LISTAGG(DISTINCT CORPUS_ITEM_ID, ',') || ']') AS CORPUS_IDS
+    "HASHED_USER_ID",
+    current_date as "UPDATED_AT",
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/user_impressions_flow.py
```diff
@@ -40,6 +40,7 @@
 
 @task
 def transform_user_impressions_df(df: pd.DataFrame) -> pd.DataFrame:
+    df = df.dropna(subset=["HASHED_USER_ID"])
```
mmiermans, line 43: Let's do this in the query above:
```sql
AND r.HASHED_USER_ID IS NOT NULL
```

mmiermans (commented):
Thanks for catching this!

Interesting that this happens.. This is probably related to an issue where [recommendation-api is getting logged-out requests](https://getpocket.atlassian.net/browse/HS-360). So far my understanding was that these recommendations would not actually reach a user, but maybe they did. At this point recommendation-api errors out when a logged-out request is made, and I don't see any more null records for `hashed_user_id` from the last 14 days.

Apart from moving the condition into the SQL query, this looks good to me.

mmiermans (approved):
:ship: :bow:

PR #186, opened 2022-11-17
Simplify html to text converstion for article content

jeshuaborges:
## Goal

Simplify html to text conversion

## Background

When performing large backfill processing jobs `html2text` was leaving artifacts in the results. It seems like we can get buy with a much more simple implementation for ML. Lets start there.

https://pocket.slack.com/archives/C03U9BMEJBC/p1668642925696229


src/common_tasks/transform_data.py
```diff
@@ -24,41 +24,33 @@ def df_field_strip(dataframe: pd.DataFrame, field_name: str, chars_to_remove: st
...
-        h.strong_mark = ''
-        h.emphasis_mark = ''
-        h.ul_item_mark = ''
 
-        re_dummy = re.compile('(.^)(.^)')
-        html2text.config.RE_MD_BACKSLASH_MATCHER = re_dummy
-        html2text.config.RE_MD_DOT_MATCHER = re_dummy
-        html2text.config.RE_MD_PLUS_MATCHER = re_dummy
-        html2text.config.RE_MD_DASH_MATCHER = re_dummy
+class MLRemover(HTMLParser):
```
eu9ene, line 28: didn't get what ML means here at first. Maybe rename it to something like MarkupRemover?

evgenyrp (approved):
looks good, I would just rename the class, up to you

src/common_tasks/transform_data.py
jeshuaborges: Suggestion applied.

evgenyrp (approved):

PR #187, opened 2022-12-15
feat(Corpus engagement): Load opens/impressions into Feature Group [TDP-181]

mmiermans:
## Goal
Load open and impression engagement stats into the Feature Group, to support [Thompson sampling](https://en.wikipedia.org/wiki/Thompson_sampling) for Home in Recommendation API.

- Engagement should by grouped by `RECOMMENDATION_SURFACE_ID`, `CORPUS_SLATE_CONFIGURATION_ID`, and `CORPUS_ITEM_ID`, such that we can get engagement metrics for the specific surface (Home) and slate (e.g. 'For You') that recommendations were shown in.
- Engagement should be available on time-scales 1, 7, 14, 21, 28 days. We’re not sure which one we’ll want to use yet, and this gives us flexibility.

## Implementation Decisions
- Key on `RECOMMENDATION_SURFACE_ID/CORPUS_SLATE_CONFIGURATION_ID/CORPUS_ITEM_ID`, to allow Thompson sampling to get Corpus Item engagement for a specific surface and slate. Also include these values separately as features to make it easy to query on them.
- Hardcode `'HOME'` as the recommendation surface id, and come back in [TDP-192](https://getpocket.atlassian.net/browse/TDP-192) to dynamically get this value from Snowflake. By dynamically getting this value we can apply Thompson sampling to surfaces other than Home, but that is out-of-scope for this epic.
- Trigger the flow every hour, because I don't (can't?) know when the Dbt model is updated.
- It is not necessary to remove stale data under the assumption that items are rarely made eligible multiple times on the same slates. In the edge-case where this does happen, Thompson sampling will use stale/historic data only for the first day, after which the feature group will be updated based on new impression and open events.

## Deployment
- [x] Create `development-corpus-engagement-v1` feature group
- [x] Create `production-corpus-engagement-v1` feature group
- [x] Merge https://github.com/Pocket/dbt-snowflake/pull/317

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/TDP-181

### DIFF
--- a/src/flows/recommendation_api/corpus_item_engagement.py
+++ b/src/flows/recommendation_api/corpus_item_engagement.py
@@ -1,0 +1,63 @@
+from prefect import task, Flow, Parameter
+from api_clients.pocket_snowflake_query import PocketSnowflakeQuery
+import pandas as pd
+import datetime
+
+from common_tasks.load_data import dataframe_to_feature_group
+from utils import config
+from utils.flow import get_flow_name, get_interval_schedule
+
+FLOW_NAME = get_flow_name(__file__)
+
+BASE_QUERY = """
+with trailing_engagement as (
+    SELECT
+        'HOME' as RECOMMENDATION_SURFACE_ID,
+        e.CORPUS_SLATE_CONFIGURATION_ID,
+        e.CORPUS_ITEM_ID,
+
+        -- aggregate impressions
+        sum(case when happened_at >= CURRENT_DATE      then impression_count else 0 end) as trailing_1_day_impressions,
+        sum(case when happened_at >= CURRENT_DATE - 6  then impression_count else 0 end) as trailing_7_day_impressions,
+        sum(case when happened_at >= CURRENT_DATE - 13 then impression_count else 0 end) as trailing_14_day_impressions,
+        sum(case when happened_at >= CURRENT_DATE - 20 then impression_count else 0 end) as trailing_21_day_impressions,
+        sum(case when happened_at >= CURRENT_DATE - 27 then impression_count else 0 end) as trailing_28_day_impressions,
+
+        -- aggregate opens
+        sum(case when happened_at >= CURRENT_DATE      then open_count else 0 end) as trailing_1_day_opens,
+        sum(case when happened_at >= CURRENT_DATE - 6  then open_count else 0 end) as trailing_7_day_opens,
+        sum(case when happened_at >= CURRENT_DATE - 13 then open_count else 0 end) as trailing_14_day_opens,
+        sum(case when happened_at >= CURRENT_DATE - 20 then open_count else 0 end) as trailing_21_day_opens,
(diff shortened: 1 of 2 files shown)


jeshuaborges (commented):
Not a blocker for this PR but something has been on my mind. I have been thinking that it may make sense to do all of our data modeling in snowflake then copy/sync those data models to feature groups. This would allow us to essentially query our features in our data warehouse and also potentially add some more conventions for syncing content to feature groups.

Something to consider...

lfishhead:
I think for clickdata, starting in snowflake limits our latency.  While initially we may never need hourly or sub-hourly updates, we may want to allow for this down the road?  

I also have a question: what is the `CORPUS_SLATE_CONFIGURATION_ID`?

src/flows/recommendation_api/corpus_item_engagement.py
```diff
@@ -0,0 +1,55 @@
...
+GROUP BY 1, 2, 3
+"""
+
+
+@task
+def transform_df(df: pd.DataFrame) -> pd.DataFrame:
+    key_columns = ['RECOMMENDATION_SURFACE_ID', 'CORPUS_SLATE_CONFIGURATION_ID', 'CORPUS_ITEM_ID']
+    df['KEY'] = df[key_columns].agg('/'.join, axis=1)
+    df['UPDATED_AT'] = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
+    return df
```
braunreyes, line 43: I seems like you can do all of this in the snowflake query, so that would reduce the need for this task at all.  Ideally you are just running a query and storing it.

braunreyes (changes requested):

mmiermans:
> I also have a question: what is the CORPUS_SLATE_CONFIGURATION_ID?

A 'CorpusSlate' is an ordered list of CorpusItem recommendations. A CorpusSlate configuration id identifies the combination of a candidate set and rankers used to produce a CorpusSlate.

src/flows/recommendation_api/corpus_item_engagement.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# TODO: Ideally, this flow would be triggered after Dbt model used in the above query is updated, instead of every hour.
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=60)) as flow:
+    snowflake_result = PocketSnowflakeQuery()(query=BASE_QUERY)
+
+    transformed_df = transform_df(snowflake_result)
+
+    dataframe_to_feature_group(transformed_df, feature_group_name=f"{config.ENVIRONMENT}-corpus-engagement-v1")
```
jeshuaborges, line 52: Do you need to wipe out all historic values? Given that some keys may no longer be receiving events but the keys will still be present and may have engagement data outside the current 27-day window.

braunreyes (approved):

PR #199, opened 2023-03-09
Flow to populate a feature group for publisher data

lfishhead:
## Goal

Publisher diversity is a consideration in both prospecting and recommendation candidate set generation.  To start to better address this need, this flow sets up a feature group that stores features indexed by `TOP_DOMAIN_NAME` that are recomputed over time.  

The features include approval data by topic from our curators if available.  This can be used to get a sense of the topical concentration or breadth of a source.  [Adding rejection data might be interesting in a future extension]

The remaining features are save based assessments of publishers over the last week:

- total saves 
- avg days between time first parsed and item save 
- avg save count (from totals)
- log total saves
- \# of items saved 

## Implementation Decisions
 
Made a feature group instead of a snowflake model to facilitate potential access in the recommendation-api.  Otherwise relatively straightforward query and  flow.

Currently set to update daily with the save aggregations

## References

[DIS-214](https://getpocket.atlassian.net/browse/DIS-214)
[DIS-295](https://getpocket.atlassian.net/browse/DIS-295)

[DIS-214]: https://getpocket.atlassian.net/browse/DIS-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/flows/prospecting/publisher_features.py
+++ b/src/flows/prospecting/publisher_features.py
@@ -1,6 +1,7 @@
 import os
 import pandas as pd
 from pathlib import Path
+from typing import Dict
 
 import prefect
 from prefect import Flow, task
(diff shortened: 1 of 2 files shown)


src/flows/prospecting/publisher_features.py
```diff
@@ -0,0 +1,77 @@
...
+    print(f"snowflake query returned {len(snowflake_result)} rows")
+
+    # transform data
+    publisher_df = transform_publisher_features.run(snowflake_result)
+
+    # load data
+    result = dataframe_to_feature_group(
+        dataframe=publisher_df,
+        feature_group_name=FEATURE_GROUP_NAME
+    )
```
braunreyes, line 74: I would suggest writing the flow like this

```suggestion
@task()
def get_snowflake_data():
    logger = prefect.context.get("logger")
    logger.info(f"Querying snowflake for publisher data")

    # Read a query from the sql/ directory located next to this flow file.
    sql_path = os.path.join(os.path.dirname(__file__), "sql/publisher_features.sql")
    sql_query = Path(sql_path).read_text()

    # create query parameters
    query_params = {"MAX_APPROVED_AGE": 360, "MIN_WORD_COUNT": 900}

    # retrieve data from snowflake
    snowflake_result = PocketSnowflakeQuery().run(
        query=sql_query,
        data=query_params,
        output_type=OutputType.DATA_FRAME,
        database=config.SNOWFLAKE_ANALYTICS_DATABASE,
        schema=config.SNOWFLAKE_ANALYTICS_DBT_SCHEMA,
    )
    logger.info(f"snowflake query returned {len(snowflake_result)} rows")
    return snowflake_result


@task()
def transform_publisher_features(input_df: pd.DataFrame) -> pd.DataFrame:

    required_columns = {
        "TOP_DOMAIN_NAME",
        "LOG_TOTAL_SAVES",
        "TOTAL_SAVE_COUNT",
        "UPDATED_AT",
    }
    if len(required_columns.difference(set(input_df.columns))) > 0:
        missing = required_columns.difference(set(input_df.columns))
        raise KeyError(f"input dataframe is missing columns: {missing}")

    # TOP_DOMAIN_NAME is required key for feature group
    input_df = input_df.dropna(subset=["TOP_DOMAIN_NAME"])
    # reformat update column
    input_df["UPDATED_AT"] = input_df.UPDATED_AT.apply(
        lambda x: x.strftime("%Y-%m-%dT%H:%M:%SZ")
    )
    # remove extra column from query
    input_df = input_df.drop(columns="APPROVED_SOURCE")

    # deal with rows for sources without approved items
    approved_cols = [c for c in input_df.columns if c.startswith("NUM_APPROVED")]
    for c in approved_cols:
        input_df[c] = input_df[c].fillna(0).astype(int)

    return input_df


with Flow(
    FLOW_NAME,
    schedule=get_interval_schedule(minutes=ONE_DAYS_MINS),
    executor=LocalDaskExecutor(),
) as flow:

    sf_data_results = get_snowflake_data()

    # transform data
    publisher_df = transform_publisher_features(sf_data_results)

    # load data
    result = dataframe_to_feature_group(
        dataframe=publisher_df, feature_group_name=FEATURE_GROUP_NAME
    )
```

braunreyes: this create a flow that looks like this when you execute with `flow.visualize()`:

```python
if __name__ == "__main__":
    # flow.run()
    flow.visualize()
```

image

braunreyes: prefect will look at your tasks added to the Flow object and execute the run methods...there is no need to call run directly.  Best practice is to create tasks and then stitch them together in the Flow object.

braunreyes (changes requested):

src/flows/prospecting/publisher_features.py
lfishhead: this is super helpful, i am still learning basic patterns let alone best practices.

braunreyes: sure no worries....Prefect v2 will make it much easier to just write Python in the flow object.  V1 gets wonky when 'untasked' execution elemets are added.

lfishhead: i also don't know how to do visualization or make use of the UI, so would be curious about that

braunreyes (approved):

PR #202, opened 2023-04-05
feat(NewTab): Load aggregate engagement to Feature Group [DIS-453]

mmiermans:
## Goal
Copy aggregate 1 day click and impression data from BigQuery to the corpus-engagement-v1 Feature Group.

## Implementation Decisions
- I tried to keep this flow as simple as possible for now. Originally I planned to load aggregated data into Snowflake. I [chose to skip this](https://getpocket.atlassian.net/browse/DIS-455) because it seems feasible to me to go directly from BigQuery to a Feature Group using Prefect. Aggregating over 1 day takes only a few seconds to [execute in ReDash](https://sql.telemetry.mozilla.org/queries/90825/source).
- We don't need the 7, 14, 21, or 28 trailing windows, so I skipped those for now.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/DIS-453

Documentation:
* [Tech spec](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2924052481/Technical+Spec+Firefox+Recommendation+API)

Dbt model with aggregate Home engagement for the corpus-engagement-v1 Feature Group:
* https://github.com/Pocket/dbt-snowflake/blob/main/models/facts/recommendation/corpus_items/corpus_recommendation_engagement_trailing_days_by_item.sql

## QA
The front-end development is still in an early phase, so there are no tileIds in BigQuery that came from Recommendation API yet. To test the flow, I used dummy data that I removed in commit https://github.com/Pocket/data-flows/pull/202/commits/0db719593350233b98c8a932e5c655417e50ee81. The following Athena query confirms that the data is loaded correctly in the Pocket-Dev Feature Group:
```sql
SELECT * FROM "sagemaker_featurestore"."development-corpus-engagement-v1-1671042691" 
WHERE recommendation_surface_id = 'NEW_TAB_EN_US';
```

### DIFF
--- a/src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
+++ b/src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
@@ -1,7 +1,7 @@
 from typing import List, Union
 
 import pandas as pd
-from prefect import Flow, task
+from prefect import Flow, task, case
 from prefect.tasks.gcp.bigquery import BigQueryTask
 
 from api_clients.pocket_snowflake_query import PocketSnowflakeQuery
@@ -21,7 +21,6 @@ EXPORT_FIREFOX_TELEMETRY_SQL = """
             AND normalized_country_code IS NOT NULL
             AND array_length(tiles) >= 1
         )
-        (
         SELECT
             CAST(flattened_tiles.id AS INT64) AS TILE_ID,
             FORMAT_DATETIME("%Y-%m-%dT%H:%M:%SZ", CURRENT_DATETIME()) as UPDATED_AT,  -- Feature Store requires ISO 8601 time format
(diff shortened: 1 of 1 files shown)


src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
```diff
@@ -0,0 +1,108 @@
...
+            0 AS TRAILING_14_DAY_OPENS,
+            0 AS TRAILING_21_DAY_IMPRESSIONS,
+            0 AS TRAILING_21_DAY_OPENS,
+            0 AS TRAILING_28_DAY_IMPRESSIONS,
+            0 AS TRAILING_28_DAY_OPENS
+        FROM impressions_data
+        CROSS JOIN UNNEST(impressions_data.tiles) AS flattened_tiles
+        GROUP BY TILE_ID
+        ORDER BY TRAILING_1_DAY_IMPRESSIONS DESC
+        LIMIT 16384 -- Limit to Snowflake's max list size. There are a lot of old items that still get some impressions.
```
mmiermans, line 42: Currently, this limit means that items with about 3 impressions/day get included.

src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
```diff
@@ -0,0 +1,108 @@
...
+        data={'tile_ids': df_column_to_list(df_telemetry, column_name='TILE_ID')},
+    )
+
+    # If none of the tileIds exist in Snowflake, then there's nothing to do. This should only happen while the new API
+    # is under development, and there are days without any NewTab impressions on content served from Recommendation API.
+    with case(df_is_empty(df_corpus_item_keys), False):
+        # Combine the BigQuery and Snowflake results on TILE_ID.
+        df_keyed_telemetry = pd_merge(df_telemetry, df_corpus_item_keys)
+        # Drop TILE_ID now we no longer need it, to match the dataframe columns with the feature group.
+        df_keyed_telemetry = df_drop(df_keyed_telemetry, columns=['TILE_ID'])
```
mmiermans, line 100: I kept Pandas functions as simple tasks for readability, instead of combining pd_merge and df_drop.

src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
```diff
@@ -0,0 +1,108 @@
...
+def df_drop(df: pd.DataFrame, columns: Union[str, List[str]]) -> pd.DataFrame:
+    return df.drop(columns=columns)
+
+
+@task()
+def df_is_empty(df: pd.DataFrame) -> bool:
+    return df.empty
+
+
+with Flow(FLOW_NAME, schedule=get_interval_schedule(minutes=15)) as flow:
```
mmiermans, line 81: I set the schedule to 15 minutes somewhat arbitrarily. This is the same frequency that FeedMachine uses today.

src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
braunreyes: I think you would know more than me on this list functionality, but is this limiting or affecting ability to deliver on acceptance criteria for data availability?  If this approach gets us what we need, then I am ok with it.

braunreyes (commented):
I left some comments...once you get a chance to answer and no changes come out of them...then this all looks good to me.

src/flows/recommendation_api/ff_new_tab_aggregated_engagement.py
mmiermans: Yes, it will only leave out some old content that still gets the occasional impression after a few months, which is no longer relevant.

braunreyes (approved):

PR #205, opened 2023-04-18
filter legacy curated candidate sets to en-US for ios next

lfishhead:
## Goal
The only client using legacy recommendation-api slates rather than corpus slates is ios-next.  The current slates don't filter by the scheduled surface which is desired.    

- This PR filters curated longreads, curated shortreads, and the collections slate to include only items that are scheduled on the en-US new tab.
- The collections slate is currently powered by metaflow and we will need an additional PR in that repo to remove the metaflow code.  This PR adds candidate generation for the collections slate to the existing curated_feeds prefect flow.

## Implementation Decisions
 
- Added candidate generation for the collections slate to curated_feeds flow
- Added scheduled surface parameter to query for longreads and shortreads flows

## References

[slack](https://pocket.slack.com/archives/C01UH57EJKD/p1680718417295649)


mmiermans (approved):
lgtm :+1:

braunreyes (approved):
LGTM as well

PR #206, opened 2023-04-19
fix(Deployment): the shell commands can be safely executed

mmiermans:
## Goal
- Values are escaped to pass them in correctly. For example, a JSON value might contain quotes that should be escaped.
- Prevent the caller from running arbitrary commands in shell.


common-utils/tests/unit/deployment/test_deployment.py
```diff
@@ -177,7 +177,7 @@ def test_flow_deployment(mock_cmd):
...
+    call_text = f"""export POCKET_PREFECT_FLOW_NAME=test.test.test && \\\n        pushd tests/unit/deployment && \\\n        prefect deployment build test_deployment.py:test_function \\\n        -n test-main \\\n        -sb github/data-flows-main/common-utils/tests/unit/deployment \\\n        -ib ecs-task/test-ECS-block \\\n        --override \'task_customizations=[{{"op": "add", "path": "/overrides/cpu", "value": "1024"}}, {{"op": "add", "path": "/overrides/memory", "value": "4096"}}, {{"op": "add", "path": "/networkConfiguration/awsvpcConfiguration/subnets", "value": ["subnet-1234", "subnet-1234"]}}, {{"op": "add", "path": "/networkConfiguration/awsvpcConfiguration/securityGroups", "value": ["sg-1234"]}}, {{"op": "add", "path": "/networkConfiguration/awsvpcConfiguration/assignPublicIp", "value": "DISABLED"}}]\' \\\n        -q prefect-v2-queue-main \\\n        -v {GIT_SHA} \\\n        --params \'{{"test_param": "test_value"}}\' \\\n        -t test -t deployment -t main \\\n        -a \\\n        --interval 120 --skip-upload && \\\n        popd"""
```
mmiermans, line 180: Changes to the expected string:
- `-interval 120` does not have quotes around `120`.
- `--override task_customizations=\'` changed to `--override \'task_customizations=`
- A space was removed after `test-ECS-block \\\n`

braunreyes (approved):

PR #215, opened 2023-05-09
feat: use max submission_timestamp for updated_at

mmiermans:
## Goal
To allow us to monitor and alert on the engagement pipeline latency, write timestamp we get from GCP to the Feature Store. 

Recommendation API will fetch this timestamp from the Feature Store and emit it to Snowplow.

## Implementation Decisions
- https://github.com/Pocket/data-flows/pull/214/commits/972e6944fedfef868961c66a418b81320ed71d0f Change is separately applied to the migration PR, but I didn't want to be blocked by that.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/DIS-586

Miro brainstorm diagram:
* https://miro.com/app/board/uXjVMT1rFfc=/


braunreyes (approved):
LGTM

PR #227, opened 2023-07-11
Fix: Fix issue to SqlJob intervals and force flows to use Cron for scheduling

braunreyes:
## Goal

1) The interval logic was not working as expected for SqlJob.  

  - The `include_now` flag was not behaving as expected.  Changes made to support the following expectations:
```
Unless include_now is True, the last interval in the list
will be the last full interval in the range.  For example, since we only
support days, if the last offset is today, there should be no intervals
to process, because the default is up through yesterday (UTC). If include_now
is True, then that final interval in the list will end before now (UTC) or the
end_date_override.
The existence of end_date_override ignores include_now = False.
```
   - Added a validation to make sure last offset passed to `get_intervals` cannot be None.
   - Updating some parameter names to make more sense.
   - Added missing tests for new interval changes
   - changing the batch start to be incremented by 1 microsecond from last_offset.  This is less aggressive than 1 millisecond.

2) Due to inconsistent behavior with `IntervalSchedules` in Prefect v2, going to force flows to use `CronSchedules`.  This is better and easier to support.  Tests, docstrings, and comments updated as needed across whole repo to support this.  the existing sql_etl deployments were also updated to use cron.

3) Due to refactor needed for impression_stats_v1, disabling schedule.
4) Added `.env.example`

## Testing

Using cron results in proper timezone aware schedules

1am PST Daily shows up as expected with the scheduled flow run times respecting my local timezone
image

removing schedule from the impression_stats_v1 deployment works as well

image

 
## References

JIRA ticket:
https://getpocket.atlassian.net/browse/DIS-837

### DIFF
--- a/common-utils/tests/deployment/unit/test_deployment.py
+++ b/common-utils/tests/deployment/unit/test_deployment.py
@@ -12,10 +12,8 @@ from common.deployment import (
     FlowDockerEnv,
     FlowEnvar,
     FlowSpec,
-    IntervalSchedule,
     PrefectProject,
     ProjectEnvs,
-    RRuleSchedule,
     get_aws_account_id,
     get_flow_folder,
     get_pyproject_metadata,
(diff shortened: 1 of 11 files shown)


common-utils/tests/deployment/unit/test_deployment.py
```diff
@@ -150,28 +148,19 @@ def test_flow_deployment(mock_cmd):
...
-    x4 = d4._get_schedule_arg()
-    assert x4 == ""
+    assert x2 == ""
 
     # Validate push deployment method using mock on run_command.
-    d5 = FlowDeployment(
+    d3 = FlowDeployment(
         deployment_name="test",
-        schedule=IntervalSchedule(interval=120),  # type: ignore
+        schedule=CronSchedule(cron="0 6 * * *", timezone="America/Los_Angeles"),  # type: ignore
```
chelseybeck, line 158: The cron schedule def seems simpler to me. Question: do we always run on Pacific vs UTC?

data-products/.env.example
chelseybeck, line 1: Thanks for adding this...should make it a little easier to setup our local env :)

common-utils/tests/deployment/unit/test_deployment.py
braunreyes: no not always depends on the people depending on the data.  Ideally we all talk in  UTC.

PR #228, opened 2023-07-14
feat: migrate new tab feature store engagement

mmiermans:
## Goal
Write aggregate click and impression counts to corpus-engagement Feature Group.

## Implementation Decisions
- Implement Feature Store ingestion asynchronously using chunks of 100, where each chunk is a Prefect task that can be retried. `put_record` is an idempotent operation, so there is no harm in retrying a chunk if it was partially successful.
  - [FeatureGroup.ingest](https://sagemaker.readthedocs.io/en/stable/api/prep_data/feature_store.html#sagemaker.feature_store.feature_group.FeatureGroup.ingest) from the sagemaker SDK was freezing on me for a long time. I'm not sure why. It's significantly slower per CPU core than asyncio.

## Todo
- [x] Add sagemaker permissions
- [x] Add FlowSpec
- [x] Test in ~staging~ dev environment

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/DIS-660


mmiermans:
@braunreyes I looked a bit into avoiding querying from BigQuery using dataframes, but in hindsight dataframes make the code a bit easier, and we'll probably want to sure them in the future as well. Might as well tackle the permission issue now.

data-products/pyproject.toml
```diff
@@ -15,6 +15,9 @@ snowflake-connector-python = "<3.0.4"
 prefect-snowflake = "^0.26.1"
 prefect-gcp = {extras = ["bigquery"], version = "^0.4.2"}
 jinja2 = "^3.1.2"
+pandas = "^2.0.3"
+db-dtypes = "^1.1.1"
+sagemaker = "^2.171.0"
```
mmiermans, line 20: I believe it took over 15 minutes to lock this file, on the latest version of Poetry. :grimacing: Some quick Google searches didn't tell me much.

mmiermans: The specific error is:

> google.api_core.exceptions.PermissionDenied: 403 request failed: the user does not have 'bigquery.readsessions.create' permission for 'projects/pocket-prefect-nonprod'

mmiermans:
@braunreyes I refactored feature_store.py to be more like a conventional Prefect flow. No more custom retry logic or semaphores, and instead each Prefect task ingests a batch of 100 rows.
- Easier to understand
- Increased performance through parallelization
- More reliable retrying

This doesn't scale to millions of rows yet because Prefect launches all tasks immediately. I don't know what the right solution is (Prefect concurrency limit? Dask configuration? Work queues?), but it's a problem we can solve later.

braunreyes (approved):
LGTM

PR #230, opened 2023-08-04
Fix/new tab disable http2

mmiermans:
## Goal
fix: Try to disable HTTP2 for new_tab_recommendations flow to see if it prevents [an error](https://app.prefect.cloud/account/20e6b12d-ffc1-4de6-b27b-c38b0bbe68f7/workspace/9ca53df3-2b17-44a6-a7d2-bdcc9d29e09c/flow-runs/flow-run/fc38cbc2-f728-4e09-bed8-4456c9492325) occurring when Prefect tries
to communicate the flow state to Prefect Cloud:

> Finished in state Failed('Flow run encountered an exception. KeyError: 175')

Suggestion coming from: https://discourse.prefect.io/t/agent-connection-closes-and-reopens/2961/2

## Implementation Decisions
- We're trying this out only for a single flow as an experiment. This is the only flow where we've observed this error, probably because it runs much more frequently than other flows.

### DIFF
--- a/data-products/pyproject.toml
+++ b/data-products/pyproject.toml
@@ -11,7 +11,7 @@ python = "^3.10"
 prefect = "^2.8.6"
 prefect-dask = "^0.2.3"
 common-utils = {path = "../common-utils", develop = true}
-snowflake-connector-python = "<3.0.4"
+snowflake-connector-python = "^3.0.0"
 prefect-snowflake = "^0.26.1"
 prefect-gcp = {extras = ["bigquery"], version = "^0.4.2"}
 jinja2 = "^3.1.2"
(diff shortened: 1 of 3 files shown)


braunreyes (approved):

data-products/pyproject.toml
```diff
@@ -11,7 +11,7 @@ python = "^3.10"
 prefect = "^2.8.6"
 prefect-dask = "^0.2.3"
 common-utils = {path = "../common-utils", develop = true}
-snowflake-connector-python = "<3.0.4"
+snowflake-connector-python = "^3.0.0"
```
mmiermans, line 14: We don't remember why this was set to < 3.0.4. Tests pass with 3.10.

PR #236, opened 2023-09-20
adding freestar reporting flow for extraction and load

chelseybeck:
## Goal

- Adding new Prefect workflow to extract Freestar data and load it into Snowflake

## Implementation Decisions
 - Instead of relying on emailed reports we decided to pull records via the API to control formatting and ensure consistency

## Deployment steps
- [ ] Database migrations?
- [ ] Secrets?

## References

JIRA ticket:
* [Link to JIRA ticket](https://mozilla-hub.atlassian.net/browse/SOCIALPLAT-230)

### DIFF
--- a/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
+++ b/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
@@ -9,14 +9,12 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 
 from prefect import flow, task
-#from prefect_snowflake.database import snowflake_query
 from prefect.server.schemas.schedules import CronSchedule
 from prefect_snowflake.database import snowflake_query_sync
 
 from common.databases.snowflake_utils import PktSnowflakeConnector
 from common.deployment import FlowDeployment, FlowEnvar, FlowSpec
 from common.settings import CommonSettings
-#from common.databases.snowflake_utils import PktSnowflakeConnector
 
 CS = CommonSettings()  # type: ignore
 
--- a/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
+++ b/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
@@ -61,10 +61,7 @@ def extract_freestar_data():
             "timeDimensions": [
                 {
                     "dimension": "NdrPrebid.record_date",
-                    "dateRange": [
-                        "2023-08-27",
-                        "2023-08-27",
-                    ],  # for backfill use "dateRange": ["2023-09-01", "2023-09-08"]
+                    "dateRange": ["Yesterday"],  # for backfill use "dateRange": ["2023-09-01", "2023-09-08"]
                 }
             ],
             "limit": 2000,  # Retrieve 2,000 records at a time to accomodate 5 second timeout
--- a/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
+++ b/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
@@ -1,0 +1,279 @@
+import requests
+import json
+import os
+import time
+import shutil
+
+import pandas as pd
+import pyarrow as pa
+import pyarrow.parquet as pq
+
+from prefect import flow, task
+#from prefect_snowflake.database import snowflake_query
+from prefect.server.schemas.schedules import CronSchedule
+from prefect_snowflake.database import snowflake_query_sync
+
+from common.databases.snowflake_utils import PktSnowflakeConnector
+from common.deployment import FlowDeployment, FlowEnvar, FlowSpec
+from common.settings import CommonSettings
+#from common.databases.snowflake_utils import PktSnowflakeConnector
+
+CS = CommonSettings()  # type: ignore
+
+# Define a subdirectory to store JSON objects
+subdirectory = "json_data"
+
+# Initialize an empty list to store the data from each JSON file
+combined_data = []
+
+# Define the output JSON file to save the combined data
+output_json_filename = "data_combined.json"
+output_json_file = os.path.join(subdirectory, output_json_filename)
+
+# Define the output Parquet file
+output_parquet_filename = "data_combined.parquet"
+
+@task(retries=3, retry_delay_seconds=5)
+def extract_freestar_data():
+    
+    # Define the API base URL and endpoint
+    API_BASE_URL = "https://analytics.pub.network"
+    API_ENDPOINT = "/cubejs-api/v1/load"
+    FREESTAR_API_KEY = json.loads(os.environ["FREESTAR_CREDENTIALS"])['api_key']
+
+    # Define the request payload with initial pagination parameters
+    request_payload = {
+        "query": {
+            "total": True,
+            "measures": [
+                "NdrPrebid.impressions",
+                "NdrPrebid.net_revenue",
+                "NdrPrebid.net_cpm"
+            ],
+            "dimensions": [
+                "NdrPrebid.record_date",
+                "NdrPrebid.network",
+                "NdrPrebid.url",
+                "NdrPrebid.utm_campaign",
+                "NdrPrebid.utm_content",
+                "NdrPrebid.utm_medium",
+                "NdrPrebid.utm_source",
+                "NdrPrebid.utm_term",
+                "NdrPrebid.site_domain",
+                "NdrPrebid.ad_unit",
+                "NdrPrebid.size",
+                "NdrPrebid.country_code",
+                "NdrPrebid.device_type",
+                "NdrPrebid.device_os",
+                "NdrPrebid.browser",
+                "NdrPrebid.integration_partner"
+            ],
+            "timeDimensions": [
+                {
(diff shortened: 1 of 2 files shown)


data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,277 @@
...
+        ),
+        FlowEnvar(
+            envar_name="FREESTAR_CREDENTIALS",
+            envar_value=f"data-flows/{CS.deployment_type}/freestar-credentials",
+        )
+    ],
+    deployments=[
+        FlowDeployment(
+            deployment_name="freestar_extraction",
+            schedule=CronSchedule(cron="0 10 * * *"),
```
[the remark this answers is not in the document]
chelseybeck: @mmiermans honestly I chose that time because we are pulling data from Freestar's "yesterday". I'm not clear on how they define "yesterday"...so was making it later in the morning. but that brings up a question I have...are prefect flows run on UTC? I assumed so...which would have this running early morning.

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,277 @@
...
+    # Loop through all JSON files in the directory
+    for filename in os.listdir(subdirectory):
+        if filename.endswith(".json"):
+            file_path = os.path.join(subdirectory, filename)
+            with open(file_path, "r") as json_file:
+                try:
+                    data = json.load(json_file)["data"]
+                    combined_data.extend(data)  # Extend the list with data blocks
+                except json.JSONDecodeError:
+                    print(f"Error reading JSON from {filename}")
```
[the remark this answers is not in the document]
chelseybeck: @mmiermans I seriously appreciate your suggestions. The thought did cross my mind to break it up into several tasks...and it makes sense to me to do that. ty

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,277 @@
...
+    # Combine paginated files
+
+    # Loop through all JSON files in the directory
+    for filename in os.listdir(subdirectory):
+        if filename.endswith(".json"):
+            file_path = os.path.join(subdirectory, filename)
+            with open(file_path, "r") as json_file:
+                try:
+                    data = json.load(json_file)["data"]
+                    combined_data.extend(data)  # Extend the list with data blocks
```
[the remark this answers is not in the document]
chelseybeck: @mmiermans ha I beat you to that and pushed those changes already :) - not sure what I was thinking in storing the files separately. i think initially it was for testing purposes and we thought it would just be one json.

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,246 @@
...
+                 $1:device_type::STRING,
+                 $1:device_os::STRING,
+                 $1:browser::STRING,
+                 $1:integration_partner::STRING,
+                 $1:impressions::INTEGER,
+                 $1:net_revenue::INTEGER,
+                 $1:net_cpm::FLOAT
+        FROM @{snowflake_table}_stage/data_combined.parquet
+        );
+"""
```
aramattamara, line 196: Also I noticed the table is created with `IF NOT EXISTS`, so I assume it might have existed already. Then some day we might inadvertently run this script twice and get duplicates in this snowflake table. Do you know is there a way to make this COPY idempotent, so it wouldn't matter if we run it multiple times, it would not create duplicates?

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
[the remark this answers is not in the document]
braunreyes: I think this is definitely something we can talk about.  For now I have created https://mozilla-hub.atlassian.net/browse/DENG-1630 as an Epic to potentially implement an orchestration flow.

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
braunreyes: Great point on the idempotent COPY

braunreyes (approved):
approving these sets of changes...will need to make one small change to stream data to file versus all in memory

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,286 @@
...
+        # Check if there are more records to retrieve
+        if record_count < 2000:
+            break  # Stop if there are no more records
+
+        # Increment the offset for the next page
+        request_payload["query"]["offset"] += 2000
+        page_number += 1
+
+        # Add a 5-second delay before fetching the next page
+        time.sleep(5)
```
gkatre, line 116: Is there a reason to add the 5 second delay?

braunreyes: @chelseybeck could you help with answering this one?

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,286 @@
...
+        df = pd.DataFrame(data)
+        df.to_parquet(output_parquet_filename.format(page_number))
+        total_count += record_count
+
+        # Check if there are more records to retrieve
+        if record_count < 2000:
+            break  # Stop if there are no more records
+
+        # Increment the offset for the next page
+        request_payload["query"]["offset"] += 2000
```
gkatre, line 112: Qs:
- will there ever be a situation where record_count > 2000?
- would it make a difference to put `+= record_count` instead of hard-coding 2000?
- would `record_count < 2000` always mean the last fetch?
- would there be an advantage to set the page size as a variable if there is a reason to make adjustments, instead of the hard-coded value of 2000?

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,286 @@
...
+format_file_sql = f"""
+CREATE OR REPLACE FILE FORMAT {snowflake_table}_format
+TYPE = parquet;
+"""
+
+create_stage_sql = f"""
+CREATE OR REPLACE STAGE {snowflake_table}_stage
+FILE_FORMAT = {snowflake_table}_format;
+"""
+
```
gkatre, line 166: Any specific advantage of adding the DDLs with the DMLs?

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
chelseybeck: Yes, it helps prevent failures due to API limits

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -0,0 +1,286 @@
...
+            create_temp_table,
+            load_sql,
+            begin_transaction,
+            delete_sql,
+            insert_sql,
+        ],
+        snowflake_connector=PktSnowflakeConnector(),
+        wait_for=[put_parquet],
+        as_transaction=True,
+    )  # type: ignore
```
gkatre, line 257: This flow looks great and easy to understand and follow! Great job!

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
gkatre: Thanks @chelseybeck! Do you think adding that context to the inline comment would be helpful?

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
chelseybeck: @gkatre the 2,000 count shouldn't change. Freestar has a 5-second timeout on their API, so we can't pull many records at a time. 2,000 was the sweet spot. I tested 2,500, 3,000, & 5,000. The higher we go the more frequently it fails. 2,000 records at a time results in the least amount of retries (I haven't seen it retry more than once, but set retries to 3 to be safe). I suppose Freestar could increase their API timeout in the future and if we were made aware we could test it all again, find the sweet spot for the number of records, and change this value.

But, you're right that it should be stored in a variable to keep things DRY. Thanks, that is a good point and I have pushed that change. :) 

Yes, record_count < 2000 always means the last fetch...I've tested this extensively - compared to manually downloading csv file for record count

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
chelseybeck: will discuss next week

data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
chelseybeck: Yes, done :)

braunreyes (approved):

PR #242, opened 2023-10-06
feat(article_text_streaming): migration of article streaming flow from v1 to v2

braunreyes:
## Goal
What changed? What is the business/product goal?

Full migration of the incremental article text flow from Prefect v1 to v2

## Implementation Decisions

Minor changes to support using Prefect v2 primitives

## Testing

Ran flow in staging using a sample of 319 files.

Here is a snapshot of the flow diagram showing the run successfully completing.

image

Here are the results:

image

This tracks because I know the test set had a bad set of html in there.

also I can see that all the test md5 hashes are in the actual production data.

image

## Deployment steps
- [N/A] Database migrations?
- [N/A] Secrets?
- [N/A] **FOR POCKET DEVELOPERS ONLY** - Post in [\#changelog](https://pocket.slack.com/archives/C0Q4UFMDZ):
> Write a #changelog message using our [best practices](https://docs.google.com/document/d/1oEt8Mtkp-6Xz9S2zaNX1EIXfQIEDN2JpY0diuPc1HZc/edit) if this PR (potentially) impacts users. Describe the 'why' and 'what'. @mention relevant teams. Link to this PR.

## References

JIRA ticket:
[Link to JIRA ticket](https://mozilla-hub.atlassian.net/browse/MSDO-6)

Documentation:
Flow documented via docstrings and comments

PR #243, opened 2023-10-06
fix(html_text_streaming): disabling the reorder while migrating to v2

braunreyes:
Disabling the reordering while waiting on migration to v2 for `parser_item_html_text_streaming_flow` flow here:

https://github.com/Pocket/data-flows/pull/242

The logic is causing the reorder to run every hour on Sunday instead of just once

## References

JIRA ticket:
* https://mozilla-hub.atlassian.net/browse/MSDO-6


gkatre (approved):
LGTM! 👍

PR #249, opened 2023-11-28
feat(aggregate_engagement): Query Glean and Activity Stream

mmiermans:
## Goal
Include Glean telemetry when ingesting aggregate clicks and impressions into the New Tab engagement Feature Group. For Firefox >= 120 can we use Glean telemetry. For Firefox < 120 we should continue to use Activity Stream.

## Implementation Decisions
I followed the existing approach of querying across BigQuery (engagement) and Snowflake (corpus item id). The limit of querying 16,384 ids is a blocker for scaling the number of recommendations, either by adding more locales or more content. This is workable for now. I foresee that we will want to increase the number of recommendations some time next year, but we don't have definitive plans yet.

Would Data Engineering have time in Q1/Q2 to make this process more scalable?
- If `STG_CORPUS_SLATE_RECOMMENDATIONS` can be ingested into BigQuery so we can do the join there. 
- Aggregating the last day of data (~300 GB) every 15 minutes isn't very efficient.
    - Dbt might incrementally aggregate click and impressions by minute.
    - [GCP Dataflow](https://cloud.google.com/dataflow?hl=en) could count clicks and impressions on streaming data using [Apache Beam](https://beam.apache.org/documentation/transforms/python/aggregation/count/).

### Duplicate events

[Counting duplicate document ids](https://github.com/Pocket/data-flows/pull/249#issuecomment-1839745927) over the last day results in:
- Total count: 560,707,507
- Duplicates: 35,122,757
- Average duplicates: 1.08

These duplicate events can be removed with [APPROX_COUNT_DISTINCT](https://cloud.google.com/bigquery/docs/reference/standard-sql/approximate_aggregate_functions) without adding much computational time:

|                       | Elapsed time | Slot time consumed |
|-----------------------|--------------|--------------------|
| No deduping           | 9 s         | 1 hr 34 min        |
| APPROX_COUNT_DISTINCT (chosen) | 8 sec       | 1 hr 48 min        |
| COUNT DISTINCT IF     | 25 s         | 2 hr 35 min        |

## Deployment steps
- [x] `firefox_desktop.newtab` must be replaced with a low latency (10-30 minutes) table before deploying this, because it updated daily, which is insufficiently fast for re-ranking New Tab based on engagement.

## References

JIRA ticket:
* [MC-124](https://mozilla-hub.atlassian.net/browse/MC-124)

Documentation:
* [pocket.click](https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/pocket_click) Glean event

### Run flow locally

1. `gcloud auth application-default login`
2. Login to the Pocket (dev) AWS account with the `PocketSSODeveloper` role
3. `cd` to the `data-products` directory
4. `poetry run python src/new_tab_recommendations/aggregate_engagement_flow.py`

Expected result:
- [x] Flows finish after a few minutes with 'completed'.
- [x] ~10 minutes later, Athena should contain the ingested engagement data. Note that `updated_at` is the time when data was ingested into BigQuery, so it's ~15 minutes earlier than when the flow was started.
```sql
SELECT count(*) FROM "sagemaker_featurestore"."development-corpus-engagement-v1-1671042691"
where updated_at >= '2023-11-29T21:45:00Z'; -- Change to 1 hour UTC time before flow was started.
``` 

image

### Glean data availability

[Querying both Activity Stream and Glean](https://sql.telemetry.mozilla.org/queries/90825/source#237681) shows that impression and click volume by day has approximately held steady when combined:

#### Impressions

image

#### Clicks

image

[MC-124]: https://mozilla-hub.atlassian.net/browse/MC-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
```diff
@@ -29,27 +30,47 @@
...
-            -- For now, we only need 1 day trailing data, so leave the other ones at 0.  
-            0 AS TRAILING_7_DAY_IMPRESSIONS,
-            0 AS TRAILING_7_DAY_OPENS,
-            0 AS TRAILING_14_DAY_IMPRESSIONS,
-            0 AS TRAILING_14_DAY_OPENS,
-            0 AS TRAILING_21_DAY_IMPRESSIONS,
-            0 AS TRAILING_21_DAY_OPENS,
-            0 AS TRAILING_28_DAY_IMPRESSIONS,
-            0 AS TRAILING_28_DAY_OPENS
+            SUM(CASE WHEN click IS NOT NULL THEN 1 ELSE 0 END) AS TRAILING_1_DAY_OPENS
```
mmiermans, line 49: 0 columns are now added in Python instead of SQL.

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
```diff
@@ -29,27 +32,47 @@
...
     """
 
+EXPORT_GLEAN_TELEMETRY_SQL = """
+    WITH events AS (
+      SELECT
+        submission_timestamp,
+        event.name AS event_name,
+        extra.value AS recommendation_id
+      FROM
+        firefox_desktop.newtab AS e
```
mmiermans: ## Blocking deployment

`firefox_desktop.newtab` needs to be changed to a live table when available. Request for one [is pending](https://mozilla.slack.com/archives/C05SWMM8W8G/p1700081099434449).

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
```diff
@@ -29,27 +32,47 @@
...
         ORDER BY TRAILING_1_DAY_IMPRESSIONS DESC
         LIMIT 16384 -- Limit to Snowflake's max list size. There are a lot of old items that still get some impressions.
     """
 
+EXPORT_GLEAN_TELEMETRY_SQL = """
+    WITH events AS (
+      SELECT
+        submission_timestamp,
+        event.name AS event_name,
+        extra.value AS recommendation_id
```
efixler: Isn't `extra` a json/map with potentially multiple keys/values? Curious about what looks like a single key/value pair here.

efixler:
> The limit of querying 16,384 ids is a blocker for scaling the number of recommendations,

I'm still a little confused here as to why every single recommended item needs its own id. It strikes me that if we're already near the bounds of a BigQuery limitation maybe we should be looking at approach to joining these with soem mechanism other than the individual recommendation id?

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
mmiermans: `extra` is coming from `CROSS JOIN UNNEST(event.extra) AS extra` below.

We're only interested in a single `extra` value (recommendation_id), so we don't have to worry about this join producing multiple rows per event. There's no risk of double-counting impressions/clicks.

mmiermans:
> > The limit of querying 16,384 ids is a blocker for scaling the number of recommendations,
> 
> I'm still a little confused here as to why every single recommended item needs its own id. It strikes me that if we're already near the bounds of a BigQuery limitation maybe we should be looking at approach to joining these with soem mechanism other than the individual recommendation id?

This is a Snowflake limitation of the number of values you can put in a SQL parameter list, like `where foobar in (1, 2, 3, ... 16384)`. Joining between databases using `where .. in` was just the simplest thing we could do, and clearly not a great long-term solution. [Braun said](https://mozilla.slack.com/archives/C05S8P70FPC/p1701375701546999?thread_ts=1701364002.590179&cid=C05S8P70FPC) that my proposals to make this more scalable are doable. When all the data is in BigQuery, we can simply join between tables, and we're no longer limited to joining at most 16,384 ids.

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
mmiermans: I simplified the query by pulling `AND extra.key = 'recommendation_id'` up to the `CROSS JOIN`:
```sql
CROSS JOIN UNNEST(event.extra) AS extra ON extra.key = 'recommendation_id'
```
This produces the same result, and hopefully it's clearer that we're querying only a single extra key.

efixler:
> When all the data is in BigQuery, we can simply join between tables, and we're no longer limited to joining at most 16,384 ids.

That sounds much better than using a long `in`!

mmiermans:
```sql
WITH events AS (
  SELECT
    e.document_id
  FROM
    `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1` AS e
  WHERE
    submission_timestamp > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 DAY)
    AND client_info.app_build >= '20231116134553' -- Fx 120 was the first build to emit recommendation_id
),

dups as (
    select document_id, count(*) as dup_cnt
    from events
    group by 1
)

select
    count(*) as total_event_count,
    countif(dup_cnt >= 2) as dup_event_count,
    avg(dup_cnt) as avg_duplicates
from dups
```
- Total count: 560,707,507
- Duplicates: 35,122,757
- Average duplicates: 1.08

mmiermans:
Using `APPROX_COUNT_DISTINCT` is slower than `COUNT DISTINCT IF`:
```sql
WITH events AS (
  SELECT
    document_id,
    submission_timestamp,
    event.name AS event_name,
    extra.value AS recommendation_id
  FROM `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1` AS e
  CROSS JOIN UNNEST(e.events) AS event
  CROSS JOIN UNNEST(event.extra) AS extra ON extra.key = 'recommendation_id'
  WHERE
    submission_timestamp > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 DAY)
    AND client_info.app_build >= '20231116134553' -- Fx 120 was the first build to emit recommendation_id
    AND event.category = 'pocket'
),
click_events as (
  select * from events where event_name = 'click'
),
impression_events as (
  select * from events where event_name = 'impression'
),
top_impressions as (
  SELECT
      recommendation_id as CORPUS_RECOMMENDATION_ID,
      FORMAT_DATETIME("%Y-%m-%dT%H:%M:%SZ", MAX(submission_timestamp)) as UPDATED_AT,  -- Feature Store requires ISO 8601 time format
      APPROX_COUNT_DISTINCT(document_id) AS TRAILING_1_DAY_IMPRESSIONS
  FROM impression_events
  GROUP BY recommendation_id
  ORDER BY TRAILING_1_DAY_IMPRESSIONS DESC
  LIMIT 16384 -- Limit to Snowflake's max list size. Can be removed once we don't join across BQ/Snowflake anymore.
)
select
  top_impressions.*,
  APPROX_COUNT_DISTINCT(click_events.document_id) AS TRAILING_1_DAY_OPENS
from
  top_impressions
join click_events on click_events.recommendation_id = top_impressions.CORPUS_RECOMMENDATION_ID
group by 1,2,3
order by TRAILING_1_DAY_IMPRESSIONS
```

chelseyklein:
> [GCP Dataflow](https://cloud.google.com/dataflow?hl=en) could count clicks and impressions on streaming data using [Apache Beam](https://beam.apache.org/documentation/transforms/python/aggregation/count/).

@braunreyes @mmiermans  fwiw...i would seriously enjoy building a beam pipeline :)

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
braunreyes: I will be implementing the live view once its ready.

braunreyes:
PR Updates:

I have fully implemented the new glean based ingestion using an incremental process that is idempotent.  The flow will populate new data from the source into a big query table called `pocket-prefect-prod.new_tab_recommendations.pocket_user_events` every 15 minutes and produce the rolling day aggregations from that table.  The flow is setup so that the first run will backfill the first days worth of data.

This PR also fully implements the worker based deployment pattern and is leveraged by this flow.  This involves adding secrets support to the `common.settings` module.  This enables the ability to easily move flows to other worker types (kubernetes, gcp cloud run, gcp vertex ai, etc.) since secrets are no longer tied to how AWS ECS does it.

This also has the changes to circleci to actually do the building and deployment.

Finally this has some minor refactoring to make class names use `Moz` instead of `Pkt` and to make tests a little bit more efficient.

Here are the results of the test in Prefect Cloud Dev

image

image

braunreyes:
new deployment also works

image

image

chelseyklein (approved):

PR #251, opened 2023-12-01
Longreads port -- with errors

efixler:
## Goal
Get the existing flows to work with Prefect v2

I made changes to `legacy_longreads_flow.py` to track the changes in `legacy_curated_flows.py` but the job is erroring.

Here is a gist file that shows the output:

https://gist.github.com/efixler/6a388614b2584de3b7faeb457389d300

NB: There are also several warnings in that output wrt name collisions. While these are warnings, they indicate naming collisions and appear to me (in particular the one around task names) as likely to materialize issues when deployed.

Ticket: https://mozilla-hub.atlassian.net/jira/software/c/projects/MC/boards/720?selectedIssue=MC-124


efixler:
The key error seems to be: 
```
10:18:02.779 | INFO    | Flow run 'banana-raven' - Submitted task run 'validate_candidate_items-0' for execution.
10:18:02.806 | ERROR   | Task run 'put_results-0' - Encountered exception during execution:
Traceback (most recent call last):
  File "/Users/ericfixler/projects/data-flows/data-products/.venv/lib/python3.10/site-packages/prefect/engine.py", line 1733, in orchestrate_task_run
    result = await call.aresult()
  File "/Users/ericfixler/projects/data-flows/data-products/.venv/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 291, in aresult
    return await asyncio.wrap_future(self.future)
  File "/Users/ericfixler/projects/data-flows/data-products/.venv/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 315, in _run_sync
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/ericfixler/projects/data-flows/data-products/src/shared/api_clients/sqs.py", line 134, in put_results
    for c in candidates:
TypeError: 'coroutine' object is not iterable
```

This _looks_ to me like perhaps there's an error getting returned here, and not a cursor/result set. If this is correct, what would be a better/more defensive way to structure this? And how can I see better what's being returned here?

efixler:
(Although `coroutine` is not something that looks at all like a result set, cursor, or error.)

data-products/src/candidate_sets/legacy_longreads_flow.py
```diff
@@ -0,0 +1,136 @@
...
+AND a.CORPUS_REVIEW_STATUS = 'recommendation'
+AND a.SCHEDULED_SURFACE_ID = %(SURFACE_ID)s
+AND a.LANGUAGE = %(LANG)s
+AND a.IS_SYNDICATED = 0
+ORDER BY REVIEWED_CORPUS_ITEM_UPDATED_AT desc
+LIMIT 90
+"""
+
+@task()
+async def transform_to_candidates(records: dict, feed_id: int) -> List[RecommendationCandidate]:
```
braunreyes: the errors are stemming from this function being async.  It does not need to be.  If we remove `async` it works as expected.

data-products/src/candidate_sets/legacy_candidate_sets_flow.py
```diff
@@ -0,0 +1,265 @@
...
+        candidate_set_ids,
+        candidate_sets,  # type: ignore
+        curated=unmapped(set_params["curated"]),  # type: ignore
+    )
+
+
+if __name__ == "__main__":
+    import asyncio
+
+    asyncio.run(create_set("topics"))  # type: ignore
```
mmiermans: Are you planning to create multiple deployments of this flow, or a flow-of-flows, to map over all `set_params_id` values?

braunreyes:
Dev Run

https://app.prefect.cloud/account/20e6b12d-ffc1-4de6-b27b-c38b0bbe68f7/workspace/9ca53df3-2b17-44a6-a7d2-bdcc9d29e09c/flow-runs/flow-run/4b5cffe0-df09-466a-aa2c-a5b43a494d5a

data-products/tests/shared/unit/test_sqs.py
```diff
@@ -0,0 +1,33 @@
...
+from prefect import flow
+from shared.api_clients.sqs import RecommendationCandidate, put_results
+
+TEST_DATA = [RecommendationCandidate(feed_id=1, publisher="test", item_id=1)]
+
+
+@mock_sqs
+@pytest.mark.parametrize("test_config", [(False, 1), (True, 0)])
+def test_sqs(test_config):
+    curated, expiration = test_config
```
mmiermans, line 13: This is how I use parametrize:
```suggestion
@pytest.mark.parametrize("curated,expiration", [(False, 1), (True, 0)])
def test_sqs(curated, expiration):
```
Just a different style. Up to you which seems better.

mmiermans (approved):
:+1: Looks good to me. :ship:

data-products/src/candidate_sets/legacy_candidate_sets_flow.py
mmiermans: I think this comment is outdated?

PR #253, opened 2023-12-07
Changing new tab extraction sources

chelseyklein:
## Goal

- New extractions are being added to replace activity stream source table with new glean data (activity stream is being retired)

## Implementation Decisions
 These changes are nearly a replica of existing jobs [1](https://github.com/Pocket/data-flows/tree/main-v2/data-products/src/sql_etl/sql/firefox_new_tab_impressions_daily) [2](https://github.com/Pocket/data-flows/tree/main-v2/data-products/src/sql_etl/sql/firefox_new_tab_impressions_hourly). The difference is that we are pulling from a different dataset `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1` - to differentiate, the new resulting tables are prefixed w/ glean_

## Deployment steps
- n/a Database migrations?
- n/a Secrets?

JIRA ticket:
https://mozilla-hub.atlassian.net/browse/MSDO-63

### DIFF
--- a/data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
+++ b/data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
@@ -9,17 +9,16 @@
 
 WITH
   deduplicated_pings AS (
-    SELECT
-      *   
-    FROM
-      `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1`
-    WHERE submission_timestamp >= {{ helpers.parse_iso8601(batch_start) }}
-    AND submission_timestamp < {{ helpers.parse_iso8601(batch_end) }}
-    QUALIFY row_number() over (PARTITION BY DATE(submission_timestamp),
-    document_id
+  SELECT
+    *
+  FROM
+    `moz-fx-data-shared-prod.firefox_desktop_live.newtab_v1`
+  WHERE
+    submission_timestamp >= {{ helpers.parse_iso8601(batch_start) }}
+    AND submission_timestamp < {{ helpers.parse_iso8601(batch_end) }} QUALIFY ROW_NUMBER() OVER (PARTITION BY DATE(submission_timestamp),
+      document_id
     ORDER BY
-    submission_timestamp desc) = 1
-)
+      submission_timestamp DESC) = 1 )
 SELECT
   DATE(submission_timestamp) AS happened_at,
   CASE
@@ -46,17 +45,17 @@ END
   COUNT(DISTINCT client_info.client_id) AS users_opening_new_tab_count,
   COUNT(DISTINCT
     CASE
-      WHEN metrics.boolean.pocket_enabled = False THEN client_info.client_id
+      WHEN metrics.boolean.pocket_enabled = FALSE THEN client_info.client_id
   END
     ) AS users_with_pocket_disabled_count,
   COUNT(DISTINCT
     CASE
-      WHEN metrics.boolean.pocket_sponsored_stories_enabled = False THEN client_info.client_id
+      WHEN metrics.boolean.pocket_sponsored_stories_enabled = FALSE THEN client_info.client_id
   END
     ) AS users_with_spocs_disabled_count,
   COUNT(DISTINCT
     CASE
-      WHEN metrics.boolean.pocket_enabled = True AND metrics.boolean.pocket_sponsored_stories_enabled = True THEN client_info.client_id
+      WHEN metrics.boolean.pocket_enabled = TRUE AND metrics.boolean.pocket_sponsored_stories_enabled = TRUE THEN client_info.client_id
   END
     ) AS users_eligible_for_spocs_count,
 FROM
@@ -66,28 +65,28 @@ WHERE
   normalized_channel = 'release'
   --include only data from Firefox 121+
   AND SAFE_CAST(SPLIT(client_info.app_display_version, '.')[0] AS int64) >= 121
-    --limit to locale & country combinations with Pocket enabled
-    AND ( ( normalized_country_code IN ('US',
-          'CA',
-          'GB',
-          'IE',
-          'IN')
-        AND metrics.string.newtab_locale IN ('en-CA',
-          'en-GB',
-          'en-US') )
-      OR ( normalized_country_code IN ('DE',
-          'CH',
-          'AT',
-          'BE')
-        AND metrics.string.newtab_locale IN ('de',
-          'de-AT',
-          'de-CH') )
-      OR (normalized_country_code IN ('IT')
-        AND metrics.string.newtab_locale IN ('it'))
-      OR (normalized_country_code IN ('FR')
-        AND metrics.string.newtab_locale IN ('fr'))
-      OR (normalized_country_code IN ('ES')
-        AND metrics.string.newtab_locale IN ('es-ES')) )
+  --limit to locale & country combinations with Pocket enabled
+  AND ( ( normalized_country_code IN ('US',
+        'CA',
+        'GB',
+        'IE',
+        'IN')
+      AND metrics.string.newtab_locale IN ('en-CA',
+        'en-GB',
+        'en-US') )
+    OR ( normalized_country_code IN ('DE',
+        'CH',
+        'AT',
+        'BE')
+      AND metrics.string.newtab_locale IN ('de',
+        'de-AT',
+        'de-CH') )
+    OR (normalized_country_code IN ('IT')
+      AND metrics.string.newtab_locale IN ('it'))
+    OR (normalized_country_code IN ('FR')
+      AND metrics.string.newtab_locale IN ('fr'))
+    OR (normalized_country_code IN ('ES')
+      AND metrics.string.newtab_locale IN ('es-ES')) )
 GROUP BY
   1,
   2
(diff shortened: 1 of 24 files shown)


data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/load.sql
```diff
@@ -0,0 +1,14 @@
+{% set sql_engine = "snowflake" %}
+{% set is_multi_statement = True %}
+{% import 'helpers.j2' as helpers with context %}
+
+{% macro table_def() %}
+    happened_at date not null,
+    feed_name string,
```
kirill-demtchouk: Should the `feed_name` column be non-null too (across all of the queries)?

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
```diff
@@ -0,0 +1,95 @@
...
+      WHEN metrics.boolean.pocket_enabled = TRUE AND metrics.boolean.pocket_sponsored_stories_enabled = TRUE THEN client_info.client_id
+  END
+    ) AS users_eligible_for_spocs_count,
+FROM
+  deduplicated_pings
+WHERE
+  --include only release channel data
+  normalized_channel = 'release'
+  --include only data from Firefox 121+
+  AND SAFE_CAST(SPLIT(client_info.app_display_version, '.')[0] AS int64) >= 121
```
kirill-demtchouk, line 67: At the same time as we make these changes, we will want to change the existing data extraction queries (that use the legacy, PingCentre tables) to filter to Firefox 120 and earlier:
```
SAFE_CAST(SPLIT(version, '.')[0] AS int64) <= 120
```

We'll want to do this for everything in these two folders:
https://github.com/Pocket/data-flows/tree/main-v2/data-products/src/sql_etl/sql/firefox_new_tab_impressions_hourly
https://github.com/Pocket/data-flows/tree/main-v2/data-products/src/sql_etl/sql/firefox_new_tab_impressions_daily

kirill-demtchouk:
I left a few comments, but otherwise this looks 🪨 solid! Thank you!

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
chelseyklein: @kirill-demtchouk i added those 6 files to this pr for review

kirill-demtchouk (approved):
🔥

braunreyes (approved):

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
```diff
@@ -0,0 +1,95 @@
...
+      'BE')
+    AND metrics.string.newtab_locale IN ('de',
+      'de-AT',
+      'de-CH') ) THEN 'NEW_TAB_DE_DE'
+    WHEN (normalized_country_code IN ('IT') AND metrics.string.newtab_locale IN ('it')) THEN 'NEW_TAB_IT_IT'
+    WHEN (normalized_country_code IN ('FR')
+    AND metrics.string.newtab_locale IN ('fr')) THEN 'NEW_TAB_FR_FR'
+    WHEN (normalized_country_code IN ('ES') AND metrics.string.newtab_locale IN ('es-ES')) THEN 'NEW_TAB_ES_ES'
+END
+  AS feed_name,
```
mmiermans, line 44: The feed name exists in Snowflake in `dbt_staging.stg_corpus_slate_recommendations.recommendation_surface_id`, to which this data can be joined using the `recommendation_id` that exists in `event.extra`. This data is reliable because it's directly based on the decision made by the recommendation API.

Maybe you could ingest the data from BigQuery without `feed_name` and then model `feed_name` in Dbt? That would remove duplicate SQL code and make `feed_name` more reliable. Given the complexity of this SQL code, I wouldn't be surprised if it doesn't perfectly match the actual feed.

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
```diff
@@ -0,0 +1,95 @@
...
+        'BE')
+      AND metrics.string.newtab_locale IN ('de',
+        'de-AT',
+        'de-CH') )
+    OR (normalized_country_code IN ('IT')
+      AND metrics.string.newtab_locale IN ('it'))
+    OR (normalized_country_code IN ('FR')
+      AND metrics.string.newtab_locale IN ('fr'))
+    OR (normalized_country_code IN ('ES')
+      AND metrics.string.newtab_locale IN ('es-ES')) )
```
mmiermans, line 89: I think we can replace all of this by looking for the presence of the `events.extras.recommendation_id`.

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_unique_engagement_by_feed/data.sql
```diff
@@ -0,0 +1,131 @@
...
+          'de-CH') )
+      OR (normalized_country_code IN ('IT')
+        AND metrics.string.newtab_locale IN ('it'))
+      OR (normalized_country_code IN ('FR')
+        AND metrics.string.newtab_locale IN ('fr'))
+      OR (normalized_country_code IN ('ES')
+        AND metrics.string.newtab_locale IN ('es-ES')) ) )
+  --let's skip the usual click deduplication step, Jeff Silverman's analysis showed that Glean is resistant to
+  --the duplicate clicks issue we saw with the legacy PingCentre events:
+  --https://docs.google.com/document/d/1aL3bjJ6PQLHH455zCMReDpYnyqPTNwTvTW2QujwP-nw/edit#heading=h.bz9b6bjjzei
```
mmiermans: I'm confused. The final query uses `COUNT DISTINCT` on `client_id` for all aggregate columns. Doesn't that mean we are deduplicating?

mmiermans (commented):
:heart: Thanks for doing this work! Love that we're pulling `recommendation_id` into Snowflake.

I'd be a bit concerned about maintainability given the amount of duplicate SQL, and SQL that's reproducing application behavior around the feed name.

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
mmiermans: I think I get it now. The problem is that we don't want to ingest that much data into Snowflake, and we don't have `stg_corpus_slate_recommendations` available in BigQuery. This makes sense again.

kirill-demtchouk: Hey Mathijs! That's right. You arrived at the conclusion yourself before I could reply 🏆  Some of these extraction queries are aggregated cuts of data without the tile ID or recommendation ID, so we can't do the mapping in Snowflake (and bringing over the raw data into Snowflake would be quite large).

On the models where we have tile/recommendation ID, we're not doing this mapping in the query itself and it will be done in dbt/Snowflake.

_Ideally_, Firefox could keep track of the feed name itself and track it directly in the Glean events!

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_unique_engagement_by_feed/data.sql
kirill-demtchouk: Yes, you're right. Sorry, I think this comment is in here by accident. @chelseybeck feel free to remove it!

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
kirill-demtchouk: That would be super clean 🤞 ...but unfortunately impressions & clicks of sponsored content won't have a recommendation_id, and we don't want to lose those events when we do all of these calculations 😞

mmiermans (approved):

data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_daily/glean_firefox_new_tab_daily_disable_rate_by_feed/data.sql
mmiermans: On second thought, this table only needs to be grouped by `recommendation_id` and `tile_id`, which is only ~700k rows per day. That could easily be loaded into Snowflake.

PR #255, opened 2023-12-15
fix: need to be able to create the dataset as part of the flow

braunreyes:
Need to have the flow create the dataset to get this deployment working

also fixing issue with main deployment dataset name

https://mozilla-hub.atlassian.net/browse/MSDO-81


braunreyes:
this PR is now converted to reverting the changes to this flow....big query access is preventing the proper dataset from being created as needed.

chelseyklein (approved):

PR #265, opened 2024-01-29
moving record count logic as it was occuring at wrong time

chelseyklein:
## Goal
The logic checking the record count was in the wrong place, causing the flow to fail
See #263 

## References

JIRA ticket:
[* Link to JIRA ticket](https://mozilla-hub.atlassian.net/browse/MSDO-62)

### DIFF
--- a/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
+++ b/data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
@@ -140,12 +140,6 @@ def extract_freestar_data(
         # Write to file
         data = response_json["data"]
 
-        record_count = len(data)
-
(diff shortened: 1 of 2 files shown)


data-products/src/freestar_revenue_reporting/freestar_reporting_flow.py
```diff
@@ -162,9 +156,15 @@ def extract_freestar_data(
             if col not in df.columns:
                 df[col] = None  # Add the column with None values
 
+        record_count = len(data)
+
         df.to_parquet(file_path.format(page_number))
```
mmiermans, line 161: :heavy_check_mark:  This function runs one more time by moving the condition down.

mmiermans (approved):

PR #266, opened 2024-02-02
feat: final flows to migrate for data-products

braunreyes:
These are the last set of flows to migrate for Prefect v2

The flow that most close to business critical would be the score update flow.

This was tested using a local version of the Pocket Web app:

image


data-products/src/update_pocket_score_data_flow.py
```diff
@@ -0,0 +1,306 @@
...
+        FlowDeployment(
+            name="deployment",
+            cron="*/60 * * * *",
+            job_variables={
+                "cpu": 2048,
+                "memory": 4096,
+            },
+        ),
+    ],
+)
```
mmiermans, line 300: Do we know what this flow is used for? Or will we just make a reasonable effort to keep it running, even if it might not be used anymore?

(To be clear, I have no idea. :grimacing: )

mmiermans (approved):

data-products/src/update_pocket_score_data_flow.py
braunreyes: yeah at this point, just going to keep it running...it does seem to be referenced in Web

braunreyes:
link to code references to tables updated by the update score flow:

https://github.com/search?q=org%3APocket%20content_quality_score_v2&type=code

@mmiermans

PR #269, opened 2024-02-07
Feature/fix dev schema deletion

braunreyes:
The dev schema clean up flow needed to be updated to be less aggressive..especially since only trying to delete schemas in the development database was not being enforced :(

This is a much better process that will only delete old tables and not the schemas.


chelseyklein (approved):
Nice!

PR #270, opened 2024-03-13
feat(legacy-candidate-sets): removing NEW_TAB_DE_DE longreads

braunreyes:
## Goal

NEW_TAB_DE_DE longreads query is no longer returning results...per @mmiermans this is ok to remove

## References

JIRA ticket:
https://mozilla-hub.atlassian.net/browse/MC-822

### DIFF
--- a/data-products/pyproject.toml
+++ b/data-products/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "data-products"
-version = "0.13.2"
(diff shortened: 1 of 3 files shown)


mmiermans (approved):
:+1: This id is not being referenced anywhere outside this repo. Recommendation-api German Discover only references candidate set id `c66a1485-6c87-4c68-b29e-e7e838465ff7`.

PR #278, opened 2024-05-14
feat(pride): create a pride slate from collections

bassrock:
## Goal

Using collection labels, turn the stories that are in them, assuming they are in the corpus, into slates. Specifically this is for Pride!

## Implementation Details

There is probably a future where all collection labels can be turned into slates, but right now we just need this one. Once we establish a rhythm and do this a few times we can look at making this more modular.


data-products/src/recommendation_api/sql/labeled_collection_stories.sql
```diff
@@ -0,0 +1,31 @@
+WITH collections_with_label_expanded AS (
```
kschelonka: don't need to do this, but this query might be slightly more obvious:

```
with labeled_stories as (
    SELECT
        slug, 
        language, 
        l.value:name as label, 
        JSON_EXTRACT_PATH_TEXT(s.value, 'url') as url
    FROM
        "ANALYTICS"."DBT"."COLLECTIONS",
        LATERAL FLATTEN(INPUT => labels) as l,
        lateral flatten(INPUT => stories) as s
    where language = 'EN'
    and label = 'idea'
)
select 
    aci.approved_corpus_item_external_id as "ID",
    aci.topic as "TOPIC",
    aci.publisher as "PUBLISHER"
from analytics.dbt.approved_corpus_items as aci
inner join labeled_stories as ls on ls.url = aci.url
where aci.approved_corpus_item_external_id != 'c931d2f5-0205-48f1-a773-dd0e682977b1'
```

bassrock: Using it!

braunreyes (approved):

mmiermans (approved):
Love that this is a simple change! :heart_hands:

PR #279, opened 2024-05-20
feat: aggregate CA New Tab engagement

mmiermans:
## Goal
[MC-1033](https://mozilla-hub.atlassian.net/browse/MC-1033) Aggregate New Tab impressions and opens from Canada, to run an experiment where we serve content to Canada that is more engaging for users there.

## Implementation Decisions
- Switch to a new table `pocket_user_events_by_country` that has a country column.
- Separate the incremental update of this table from 

## Deployment steps
- [ ] Do we need to do anything to create `pocket_user_events_by_country`?
- [ ] Delete `pocket_user_events`?

## References

Documentation:
* [Experimentation brief](https://docs.google.com/document/d/14SfGmWPwnzWv7oeDsHXZuRT-EUISEGqXCQmHRaoP6Lg/edit)

## QA

- [x] Push to dev-v2
- [x] Trigger flow

Query aggregate opens and impressions written to the Feature Group for CA:
```sql
SELECT
    SUM(trailing_1_day_opens) as total_open_count,
    SUM(trailing_1_day_impressions) as total_impression_count
FROM "sagemaker_featurestore"."development-corpus-engagement-v1-1671042691"
where key like '%/CA';
```

Query [aggregate opens and impressions from BigQuery for CA](https://sql.telemetry.mozilla.org/queries/100050/source).

Expectation:
- [x] Counts match approximately.

[MC-1033]: https://mozilla-hub.atlassian.net/browse/MC-1033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
```diff
@@ -189,16 +222,27 @@ async def export_telemetry_by_corpus_item_id(
...
+async def aggregate_engagement(region: str = None):
     """
     Ingests NewTab telemetry joined with Corpus metadata into a Sagemaker Feature Group.
     """
     logger = get_run_logger()
+
+    # Do an incremental update of the events table.
+    await bigquery_query(
+        gcp_credentials=MozGcp(), query=UPDATE_POCKET_USER_EVENTS_BY_COUNTRY_SQL
+    )
```
mmiermans, line 234: This query should run before the ones below. Is it still necessary to define a dependency on those tasks in Prefect v2?

If so, would the cleanest way to do that be to add `if result is not None:`?

data-products/src/new_tab_recommendations/aggregate_engagement_flow.py
```diff
@@ -189,16 +222,27 @@ async def export_telemetry_by_corpus_item_id(
 
 
 @flow(name="new-tab-recommendations.aggregate-engagement")
-async def aggregate_engagement():
+async def aggregate_engagement(region: str = None):
```
mmiermans: Remove unused parameter:
```suggestion
async def aggregate_engagement():
```

braunreyes (approved):

PR #282, opened 2024-05-28
fix(labeled-collections): ensuring that the stories in collection slates always have a topic and publisher

bassrock:
## Goal

Make it so the corpus slate job does not fail if a topic or publisher is missing by not returning those items.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -250,7 +250,7 @@ jobs:
                 condition: << parameters.use_deploy_cli >>
                 steps:
                   - setup_remote_docker:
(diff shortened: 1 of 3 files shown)


.circleci/workflows.yml
```diff
@@ -250,7 +250,7 @@ jobs:
                 condition: << parameters.use_deploy_cli >>
                 steps:
                   - setup_remote_docker:
-                      version: 20.10.18
+                      version: docker24
```
bassrock, line 253: Circleci is deprecating the older docker versions

braunreyes (approved):

PR #284, opened 2024-06-18
fix(article_streaming): updating the delimiter and escaping logic to prevent errors

braunreyes:
Error is pasted into the JIRA ticket.  Changing the delimiter to '|' for less chance of conflict.  Also adding proper '\' escaping for csv file upload.

## References

JIRA ticket:
https://mozilla-hub.atlassian.net/browse/DENG-4086


chelseyklein (approved):
🙌

PR #287, opened 2024-06-24
fix: apply timezone filtering to android recs

jpetto:
## The Problem

due to timezone issues, android home is displaying items before they should be seen, and likely before curators have had a chance to look at them (for ML scheduled items). 

## Goal

apply timezone filtering to both the reviewed items query and the scheduled items query. the intent of the fix is to only return items that have been approved and/or scheduled before the current timestamp in EST minus 3 hours. 

this script is only ever called with `NEW_TAB_EN_US` as the surface, and (afaik!) is only used to generate slates for android home recommendations.

- remove 'MANUAL' restriction on corpus items (not needed now that we have timezone filtering)

## Implementation Decisions
 
hard-code for `NEW_TAB_EN_US` as that's the only surface against which this script is run (as far as i can tell - see the config [here](https://github.com/Pocket/data-flows/blob/main-v2/data-products/src/shared/models/corpus_candidate_set_configs.py#L71-L76)).

## References

JIRA ticket:
* https://mozilla-hub.atlassian.net/browse/MC-1239

### DIFF
--- a/data-products/pyproject.toml
+++ b/data-products/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "data-products"
-version = "0.14.15"
+version = "0.14.16"
 description = "Main project folder for data products prefect flows"
 authors = ["Braun <breyes@mozilla.com>"]
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


mmiermans (approved):
Good thinking! 🧠

lfishhead (approved):
This looks fine to me.  If we wanted to further simplify, I think we need to only grab from `scheduled_corpus_items` since we are working at higher volume and very few items are in the corpus that aren't scheduled these days for en-US anyway.  But this is basically a nit.

gkatre (approved):
@jpetto Looks good!

I also think if the `recently_updated_items` cte does not add much value (that is: most topics end up having at least one schedule), then perhaps consider dropping it?

PR #288, opened 2024-06-25
refactor(sql_etl): cleaning remaining deprecated macro usage and fixing cleanup

braunreyes:
There were more usages of the deprecated macro that needed to be cleaned up.

https://mozilla-hub.atlassian.net/browse/AD-324

also the changes to the hourly jobs were cause a lot of files to get deleted, so I needed to update the cleanup to directly delete via the GCP SDK.  If the backfill intervals are still taking a long time than I will work with DATA SRE to clean out these folders.

I also updated python dependencies due to dependabot alerts

`common-utils/src/common/cloud/gcp_utils.py`

added the ability to reference environment specific gcs bucket to use

`data-products/src/shared/utils.py`

created special task to handles the file cleanup

`data-products/src/sql_etl/run_jobs_flow.py`

added better subflow naming and implement the 2 changes above into the flows

The sql changes are all to make sure the the date filters work as expected.

`data-products/src/sql_etl/sql/extras/helpers.j2`

The previous `>=` was overly destructive.  Since we are only needing to support DATE level aggregations, simply changing this to '=' ensures that only the interval being process will be merged with out deleting everything after.

Tested by running the hourly for current date, which added data for 6/26 and 6/27

image

Then I ran for 5/1 - 5/5 to make sure that a backfill would not kill the 6/26 and 6/27 data.

image

image


chelseyklein (approved):
Nice 🙌 - changes look good

PR #289, opened 2024-06-27
fix(sql): sql statements that are used for loading data into snowflake from BQ need to be protected from tile_ids as strings

braunreyes:
`data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_hourly/glean_firefox_new_tab_daily_engagement_by_tile_id_position/data.sql` and 
`data-products/src/sql_etl/sql/glean_firefox_new_tab_impressions_hourly/glean_firefox_new_tab_daily_engagement_by_tile_id_position_country_locale/data.sql`

will safe_cast on tile_id and set bad ones to -1, just like we did with position

`data-products/src/new_tab_recommendations/aggregate_engagement_flow.py`

this will filter out tile_ids that are not ints

https://mozilla-hub.atlassian.net/browse/AD-426

### DIFF
--- a/data-products/pyproject.toml
+++ b/data-products/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "data-products"
-version = "0.14.14"
+version = "0.14.15"
 description = "Main project folder for data products prefect flows"
 authors = ["Braun <breyes@mozilla.com>"]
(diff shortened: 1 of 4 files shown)


chelseyklein (approved):
🙌

PR #290, opened 2024-06-28
refactor(sql_etl): reverting new file delete logic because it is causing sporadic flow errors

braunreyes:
For some reason the GCP SDK is throwing errors about not being able to delete files because they do not exist.  However, they do exist and on retry it works.  Since this never happened with the old logic and testing shows no improvement in deletion speed.  I am reverting this change.

This is still related to work on https://mozilla-hub.atlassian.net/browse/AD-324


chelseyklein (commented):
makes sense to revert

chelseyklein (approved):

PR #292, opened 2024-07-01
refactor(common-utils): removing old deployment logic and fixing recommended_by_recency

braunreyes:
The old deployment logic was causing flows to be double deployed.  We no longer run agent based flow executions, so that logic is no longer needed.

I have also added to fixes to support recommended_by_recency SQL for corpus_candidate_sets:

- added support for multi statement SQL
- added a where clause to recommended_by_recency to omit records where TOPIC is null


data-products/src/recommendation_api/sql/recommended_by_recency.sql
```diff
@@ -58,4 +58,5 @@ SELECT
     TOPIC,
     PUBLISHER
 FROM deduped
+WHERE TOPIC is not null
```
mmiermans, line 61: Might as well apply this to `publisher`?
```suggestion
WHERE TOPIC is not null
AND PUBLISHER is not null
```

data-products/src/recommendation_api/corpus_candidate_sets_flow.py
```diff
@@ -93,12 +93,23 @@ async def load_corpus_candidate_set_records(
...
+    if candidate_set_config.is_multiquery:
+        # if multiquery we need to use different function
+        sql_split = sql_query.split(";")
+        results = await snowflake_multiquery(
+            queries=sql_split[:-1],  # omit last item which will be blank
+            snowflake_connector=snowflake_connector,
+            params=candidate_set_config.query_params,  # type: ignore
+            cursor_type=DictCursor,  # type: ignore
+        )
+        corpus_items = results[-1::][0]  # only need the last set of results
```
mmiermans, line 105: I appreciate how you often go for the clean, long-term solution.

If [recommended_by_recency.sql](https://github.com/Pocket/data-flows/pull/292/files#diff-188289ca819d7b78d103190dbaff850db48f73166a7920d8a8f6dc2dd45be550) is the only case where we need multiqueries, I would've simply refactored that SQL into a single query. For example by duplicating `max_timestamp` in-place, or by adding a `with max_timestamp as ( ... )` to join against.

mmiermans (commented):
Thanks for fixing this! 🙇

jpetto (approved):
not too familiar with the CI stuff (but i don't see anything obviously wrong).

thank you for cleaning up my mess introduced by the multiquery thing (i didn't even know this was a thing in snowflake - clearly 😅).

mmiermans (approved):

data-products/src/recommendation_api/corpus_candidate_sets_flow.py
braunreyes: Thank You

PR #294, opened 2024-07-10
refactor(data_products): deprecating sql_etl in favor of new firefox_newtable flow

braunreyes:
This is the creation of a brand new flow called 'firefox_newtab` that replaces the newtab extract and load that was managed via the `sql_etl` flow.  The 'sql_etl` was way too complex for only needed to execute newtab extractions.

This deletes the `sql_etl` code as part of this refactoring

Also updated deployment to use static flow names to prevent double deployment that was happening.

Here is the ticket: https://mozilla-hub.atlassian.net/browse/AD-427

Test Results from a daily and hourly run in staging: https://mozilla-hub.atlassian.net/wiki/spaces/PE/pages/862191667/Firefox+ETL+Refactor+Test+Results


chelseyklein (approved):
Thanks @braunreyes this is much simpler 😄

PR #302, opened 2024-09-13
Deng 4478 filter out null topics

chelseyklein:
## Goal

- Updated SQL to join to `corpus_items_updated` in order to filter out null topics that were causing a validation failure

## References

JIRA ticket:
* [Link to JIRA ticket](https://mozilla-hub.atlassian.net/browse/DENG-4478)


data-products/src/recommendation_api/sql/scheduled_syndicated.sql
```diff
@@ -1,11 +1,13 @@
...
-ORDER BY SCHEDULED_CORPUS_ITEM_SCHEDULED_AT DESC
+    s.APPROVED_CORPUS_ITEM_EXTERNAL_ID as "ID",
+    i.TOPIC as "TOPIC",
+    i.PUBLISHER AS "PUBLISHER"
+FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS" AS s
+LEFT JOIN "ANALYTICS"."DBT"."CORPUS_ITEMS_UPDATED" AS i
+ON s.APPROVED_CORPUS_ITEM_EXTERNAL_ID = i.APPROVED_CORPUS_ITEM_EXTERNAL_ID
+WHERE s.SCHEDULED_SURFACE_ID = %(SCHEDULED_SURFACE_ID)s
+AND i.TOPIC IS NOT NULL
+AND s.IS_SYNDICATED
```
gkatre:
```suggestion
AND i.IS_SYNDICATED
```
`IS_SYNDICATED` happens to be an attribute of corpus item. This way we are picking the most recent update (if any)

gkatre (approved):
@chelseybeck just one suggestion...otherwise looks good!
👍

data-products/src/recommendation_api/sql/scheduled_syndicated.sql
```diff
@@ -1,11 +1,13 @@
...
-AND approved_corpus_item_external_id NOT IN ('904bc245-54de-412a-a8af-c13eb0a7c15c') -- See #incidents on 2023-03-21 and DENG-4478
-AND IS_SYNDICATED
-AND SCHEDULED_CORPUS_ITEM_SCHEDULED_AT BETWEEN DATEADD(day, %(MAX_AGE_DAYS)s, CURRENT_DATE) AND CURRENT_DATE
-QUALIFY row_number() OVER (PARTITION BY APPROVED_CORPUS_ITEM_EXTERNAL_ID ORDER BY SCHEDULED_CORPUS_ITEM_SCHEDULED_AT DESC) = 1
-ORDER BY SCHEDULED_CORPUS_ITEM_SCHEDULED_AT DESC
+    s.APPROVED_CORPUS_ITEM_EXTERNAL_ID as "ID",
+    i.TOPIC as "TOPIC",
+    i.PUBLISHER AS "PUBLISHER"
+FROM "ANALYTICS"."DBT"."SCHEDULED_CORPUS_ITEMS" AS s
+LEFT JOIN "ANALYTICS"."DBT"."CORPUS_ITEMS_UPDATED" AS i
```
gkatre, line 6: fyi: does not make a big diff if its an outer join vs. an inner join here. the `TOPIC IS NOT NULL` will force an inner join anyway.

Also, want to take this opportunity to answer your previous question of why I chose -3 hours for `corpus_item_schedules_updated`. Its because _sometimes_ a newly added corpus item (although created when a schedule gets created) could come in with a short delay (Snowplow latency). The -3 could address that situation by reprocessing the schedules again and backfilling with the previously missing metadata. Note: this approach is not meant to be used for other reasons (example: for situations when curators do not end up adding topic data for the corpus items)

```suggestion
JOIN "ANALYTICS"."DBT"."CORPUS_ITEMS_UPDATED" AS i
```

mmiermans: Why 3 hours?

chelseyklein: this was merged prior to the join suggestion above @gkatre so if it doesn't matter then i'll leave it as is

gkatre: @mmiermans you seem to have asked the same question on slack. Replied [there](https://mozilla.slack.com/archives/C05EVRU1U1X/p1726263224023009?thread_ts=1725910997.253949&cid=C05EVRU1U1X)

PR #306, opened 2025-01-15
Deng 7246 add UAPI data to newtab reporting

chelseyklein:
## Goal

- Queries BigQuery and creates a new table in Snowflake, `prefect.ads_daily_engagement_by_tile_id_position`, to bring ads backend impression data into Snowflake 

We will also be updating [this dbt model](https://github.com/Pocket/dbt-snowflake/blob/main/models/facts/firefox_new_tab/firefox_new_tab_daily_engagement_by_tile_id_position_x.sql) to union this data on existing newtab impression data that feeds Mode reporting

## References

JIRA ticket:
[DENG-7246](https://mozilla-hub.atlassian.net/browse/DENG-7246)

- https://github.com/Pocket/dbt-snowflake/pull/541

[DENG-7246]: https://mozilla-hub.atlassian.net/browse/DENG-7246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data-products/src/firefox_newtab/sql/ads_impressions_hourly/ads_daily_engagement_by_tile_id_position/data.sql
```diff
@@ -0,0 +1,27 @@
...
+      ELSE 0
+  END
+    ) AS click_count,
+  0 AS save_count,
+  0 AS dismiss_count
+FROM
+  ads_backend.interaction AS i
+WHERE 
+    submission_timestamp >= '{{ batch_start }}'
+    AND submission_timestamp < '{{ batch_end }}'
```
curtismorales: We'll need to do the same logic for handling live vs stable tables as exists in the other job https://github.com/Pocket/data-flows/blob/main-v2/data-products/src/firefox_newtab/sql/glean_firefox_new_tab_impressions_hourly/glean_firefox_new_tab_daily_engagement_by_tile_id_position/data.sql#L4-L24

We also need to do some filtering to SPOC desktop interactions

I think this will do it, I'll do some testing as well to get the query working

```suggestion
FROM
  {% if with_stable %}
  `moz-fx-data-shared-prod.ads_backend_stable.interaction_v1` AS i
  {% else %}
  `moz-fx-data-shared-prod.ads_backend_live.interaction_v1` AS i
  {% endif %}
WHERE 
  submission_timestamp >= '{{ batch_start }}'
  AND submission_timestamp < '{{ batch_end }}'
  AND i.metrics.string.ad_client_form_factor = 'desktop'
  AND i.metrics.string.ad_client_placement = 'newtab_spocs'
QUALIFY 
  ROW_NUMBER() OVER (
    PARTITION BY DATE(submission_timestamp), document_id
    ORDER BY submission_timestamp DESC
  ) = 1
```

I'm not sure if we can do this and the GROUP BY, we might need to put all the logic except the aggregation into a CTE and then do the GROUP BY

curtismorales (changes requested):

curtismorales:
One other thought -- maybe we should call this `uapi_interactions_hourly` instead of `ads_impressions_hourly`?

data-products/src/firefox_newtab/sql/ads_impressions_hourly/ads_daily_engagement_by_tile_id_position/data.sql
```diff
@@ -0,0 +1,27 @@
+SELECT
+  DATE(i.submission_timestamp) AS happened_at,
+  NULL AS recommendation_id,
+  COALESCE(SAFE_CAST(i.metrics.string.ad_id AS int), -1) AS tile_id,
```
curtismorales: Let's also call this `ad_id`, we don't need to change the name here
```suggestion
  COALESCE(SAFE_CAST(i.metrics.string.ad_id AS int), -1) AS ad_id,
```

chelseyklein:
> One other thought -- maybe we should call this `uapi_interactions_hourly` instead of `ads_impressions_hourly`?

i thought about this too...but since it comes from `ads_backend` i called it `ads`...can change

data-products/src/firefox_newtab/sql/ads_impressions_hourly/ads_daily_engagement_by_tile_id_position/data.sql
curtismorales: Dropping in an updated suggestion, closing this one

data-products/src/firefox_newtab/sql/ads_impressions_hourly/ads_daily_engagement_by_tile_id_position/data.sql
```diff
@@ -0,0 +1,27 @@
...
+FROM
+  ads_backend.interaction AS i
+WHERE 
+    submission_timestamp >= '{{ batch_start }}'
+    AND submission_timestamp < '{{ batch_end }}'
+GROUP BY
+  1,
+  2,
+  3,
+  4
```
curtismorales: I needed to split the first bit into a CTE to make it work, otherwise this is mostly the same

```suggestion
WITH uapi_interactions_deduped AS (
  SELECT
    DATE(submission_timestamp) AS happened_at,
    NULL AS recommendation_id,
    COALESCE(SAFE_CAST(metrics.string.ad_id AS int), -1) AS ad_id,
    COALESCE(SAFE_CAST(metrics.string.ad_client_position AS int), -1) AS position,
    metrics.string.ad_interaction AS interaction_type,
  FROM
    {% if with_stable %}
    `moz-fx-data-shared-prod.ads_backend_stable.interaction_v1` AS i
    {% else %}
    `moz-fx-data-shared-prod.ads_backend_live.interaction_v1` AS i
    {% endif %}
  WHERE 
    submission_timestamp >= '{{ batch_start }}'
    AND submission_timestamp < '{{ batch_end }}'
    AND metrics.string.ad_client_form_factor = 'desktop'
    AND metrics.string.ad_client_placement = 'newtab_spocs'
  QUALIFY 
    ROW_NUMBER() OVER (
      PARTITION BY DATE(submission_timestamp), document_id
      ORDER BY submission_timestamp DESC
  ) = 1
)
SELECT
  happened_at,
  recommendation_id,
  ad_id,
  position,
  SUM(
    CASE
      WHEN interaction_type = 'impression' THEN 1
      ELSE 0
    END
  ) AS impression_count,
  SUM(
    CASE
      WHEN interaction_type = 'click' THEN 1
      ELSE 0
    END
  ) AS click_count,
  0 AS save_count,
  0 AS dismiss_count,
FROM uapi_interactions_deduped
GROUP BY
  1,
  2,
  3,
  4
```

curtismorales (approved):
Looks good to me!

